PR #9, opened 2015-02-11
Allow admin users to create/update/destroy offices

stephenrichards:
The office controller specs still need to be done to prove what actions admin and non-admin users can do.


colinbruce:
Code review passed

PR #12, opened 2015-02-17
User invitations 

colinbruce:
Allows admins to invite users, with enhanced specs


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,36 @@
+class UsersController < ApplicationController
+  respond_to :html
+  before_action :authenticate_user!
+  load_and_authorize_resource
+
+  def index
+    @users = User.all.order('lower(email)')
```
stephenrichards: might be better to write a scope :all_by_email on the User model, and then here do User.all_by_email

stephenrichards:
Code review: one comment regarding re-writing the users find all order by email as a scope, but not esssential at this point, :+1:

PR #13, opened 2015-02-17
Minor updates

colinbruce:
Improved scope for users and rendering in specs


dotemacs:
Looks good. Really like that the specs have been tidied up :)

PR #16, opened 2015-02-20
Implement rubocop

colinbruce:
Added gems, setup a rubocop instruction file and then implemented all suggested changes


colinbruce:
Do not merge yet.  More changes will be needed after DWP checker is merged in.

.rubocop.yml
```diff
@@ -0,0 +1,138 @@
...
+  # generated code and copy-and-pasted snipets from READMEs.
+  Exclude:
+    - 'Guardfile'
+    - 'vendor/**/*'
+    - 'db/**/*'
+    - 'bin/**/*'
+    - '**/*.gemspec'
+    - '**/Rakefile'
+    - '**/Vagrantfile'
+    - 'spec/rails_helper.rb'
```
dotemacs: I wouldn't exculde rails_helper.rb as it'll hold configuration but it'll still be code.

Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby '2.1.5'
```
dotemacs: I wonder if we should go to 2.2.0 here as it's already out...

Guardfile
```diff
@@ -1,43 +1,75 @@
...
+    # Rails files
+    rails = dsl.rails(view_extensions: %w(erb haml slim))
+    dsl.watch_spec_files_for(rails.app_files)
+    dsl.watch_spec_files_for(rails.views)
+
+    watch(rails.controllers) do |m|
+      [
+          rspec.spec.("routing/#{m[1]}_routing"),
+          rspec.spec.("controllers/#{m[1]}_controller"),
+          rspec.spec.("acceptance/#{m[1]}")
```
dotemacs: This line and the two above are all indented by 4 spaces :)

dotemacs:
Apart from the two comments above, which aren't show stoppers, it seems fine to me.

PR #21, opened 2015-02-26
improved handling of user tests and editing

colinbruce:
Implemented a user edit function and improved test coverage of the user_controller

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -94,14 +94,12 @@ RSpec.describe UsersController, type: :controller do
     end
     describe 'GET #show' do
       it 'shows user details' do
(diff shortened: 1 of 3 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -82,5 +97,55 @@
...
+      context 'with valid params' do
+        let(:new_attributes) {
+          {
+            email: 'new_attributes@example.com',
+            password: 'aabbccdd',
+            role: 'user'
+          }
+        }
+
+        it 'updates the requested user' do
```
dotemacs: Can't you just declare in a let block & re-use the user instance across the examples below?

dotemacs:
Looks good :+1:

PR #32, opened 2015-03-27
Fix the call to DWP checker

mattmb:
This view included a call to a previous method, it now calls the correct function.


mattmb:
@dotemacs @colinbruce the fix we discussed.

@colinbruce will be pleased to here the E2E connection works :+1:

colinbruce:
Approved and merged :+1:

PR #35, opened 2015-03-30
Configure logging

colinbruce:
Enable and configure the production logging


dotemacs:
If this is the approach we've taken previously, then merge away

PR #45, opened 2015-04-17
Build assets in container

mattmb:
Our build process compiles the assets in the container, then exports them and uploads to S3 during deployment.


colinbruce:
:+1: Merge at will

PR #46, opened 2015-04-17
Allow Admins to assign Users to an Office

dotemacs:
Every user should be assigned to an office. The idea is to have an admin
per office so that when that admin creates/invites a user, that user
inherits that admin's office assignment.


colinbruce:
Looks good, merge with pride :+1:

PR #47, opened 2015-04-17
Income calculator

colinbruce:
Add the income calculator (R2) with shortcut on toolbar

Requirement was for an on-page JS solution that did not need
postback to server.

This solution uses - 
* Coffeescript called from calculator page to generate result 
* Uses Ajax to send it to the DB for storage/stats generation


dotemacs:
After discussing it with @colinbruce I'm happy to approve this PR.

PR #51, opened 2015-04-22
(WIP) Vagrant with postgres

mattmb:
To make it easier for people to pickup this project I have added postgres to the vagrant box.

Something else that would be good is to add some seed data for an admin user so that people can login.
@colinbruce can we add that seed data? Is it easy to make sure it doesn't run in production?


colinbruce:
@mattmb Your vagrantfile update worked fine for me.  Added user seed data.  Happy to merge?

PR #58, opened 2015-04-27
Fix for r2 calc

colinbruce:
* moved calculations out of conditional block
* and placed parsed json on page in case of error


colinbruce:
@dotemacs Alex.  I found a slight bug in the calculator.
Can you review please?

dotemacs:
Looks fine

PR #61, opened 2015-04-29
Added white listed email domains

dotemacs:
So that nobody outside of those domains can be invited to the
application.
Right now it's only allowing: (hmcts.gsi|digital.justice).gov.uk.


colinbruce:
Looks good to go, once the tests pass :+1:

PR #62, opened 2015-04-29
New - Create unique identifier for API

colinbruce:
The front end should generate a unique value in the format:

username
timestamp in YYMMDDHHSS
unique id
Which will look like bsmith@YYMDDHHSS.some-id. 
It should be up to 50 characters long, as that is the constraint imposed by DWP API.


dotemacs:
I'll merge it as soon as the build goes green

PR #67, opened 2015-05-06
Style error messages

colinbruce:
* Set text to use en locale file as per content doc
* Display errors as per design
* Updated tests to reference changes and use locale file


dotemacs:
:+1:

PR #68, opened 2015-05-06
Added error message for non white listed emails

dotemacs:
Currently we should only allow emails from white listed domains. This
change introduces meaningful and descriptive error messages, so that
when the admin users add a non MOJ approved email, they will know what
the issue is.

The ticket for this PR is [here](https://www.pivotaltracker.com/story/show/93001180).


colinbruce:
Looks good, as long as travis goes green!

PR #70, opened 2015-05-07
Error message update

colinbruce:
Updated styling in line with [DWP // Global // Error messaging ticket](https://www.pivotaltracker.com/story/show/93654782)

* Created validation messages and labels in locale file

Updated user screens
* moved to foundation base
* updated text and labels inline with copy changes
* moved validation and display texts into locale file
* updated tests to reflect text changes


dotemacs:
Looks good. I like what you've done for the email validation & error message!

PR #88, opened 2015-05-22
Update DWP availability error

colinbruce:
Refactored the dwp_process service

* it now has a result method that returns a json object with three attributes
    - success
    - dwp_check
    - message

* The success should be true or false
* The message is any error that can be trapped on return from
the downstream APIs
* The dwp_check is the serialised object

[finishes #94226788]


dotemacs:
Looks fine

dotemacs:
Looks good

PR #90, opened 2015-05-26
94407712 date validation error on dwp check form

dotemacs:
Prevent the error popping up when an invalid date is entered for the date of birth of for the date_application_made.

Also added constraints such as 16 years being the minimum age and 120 years being the maximum age of an applicant. The logic is that we've agreed as a team to have 120 be the maximum age. For minimum age, the reason it's 16 is because you get your national insurance card at that age.


colinbruce:
Looks good to me :+1:

PR #93, opened 2015-05-27
Crest style update

colinbruce:
The crest should be indented when
screen is at medium and lower


colinbruce:
Can be tested at https://fr-staff-app.herokuapp.com/feedback

kelliematheson:
Tested on Heroku, crest is now indented, happy to merge

PR #102, opened 2015-06-01
Admin graph update 

colinbruce:
Admins see R1 graphs for each office
Updated controller logic to build hash with office name and 
dwp_check count by type data for each, non-digital, office.
[finishes #94093060]

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -2,6 +2,7 @@ class Office < ActiveRecord::Base
   has_many :users
 
   scope :sorted, -> {  all.order(:name) }
+  scope :non_digital, -> { where('name != ?', 'Digital') }
 
   validates :name, presence: true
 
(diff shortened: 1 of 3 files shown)


app/models/office.rb
```diff
@@ -2,6 +2,7 @@ class Office < ActiveRecord::Base
   has_many :users
 
   scope :sorted, -> {  all.order(:name) }
+  scope :non_digital, -> { where('name != ?', 'Digital') }
```
dotemacs: PEDANTIC HAT ON.

A style suggestion:

```ruby
where.not(name: 'Digital')
```

Reference: https://robots.thoughtbot.com/activerecords-wherenot

PEDANTIC HAT OFF.

dotemacs:
Hello @colinbruce 

it looks good.

I just made an observation about a different scope style, see what you think of it. But it's not a merge blocker.

Thanks :smile:

colinbruce:
:+1: for the review and :+1: for the tip.  I much prefer that syntax!

colinbruce:
[delivers #94093060] on demo

PR #107, opened 2015-06-05
Create manager role

colinbruce:
Allow admins to create managers per-office

A manager can:
* invite users and managers to the same office as themselves
* edit users from their own office, fields:
  - name, email, role and office
* see a list of DWP checks on their dashboard that have 
been undertaken in their office
* they cannot see or amend users in other offices, or admins 
within their own office

**Note** : When a user is moved into another office, the manager
is shown the email addresses for managers in the new office

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ gem 'logstasher'
 
 # Date validation
 gem 'date_validator'
+gem 'will_paginate'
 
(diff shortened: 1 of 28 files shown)


colinbruce:
[finishes #93738888]
[finishes #95978294]
[finishes #96090702]

PR #111, opened 2015-06-09
Create new admin graph

colinbruce:
This creates a new graph on the admin page
It shows checks performed against time, 
this allows us to see times that are optimal for
publishing updates.

It also increases test coverage of the index view 
for different roles.


dotemacs:
Looks fine

PR #112, opened 2015-06-09
Improving tests

colinbruce:
First pass at standardising test layout

Replaced the fixtures setting with FactoryGirl 
in the rails helper config.

Removed all explicit FactoryGirl.* calls

Tried to standardise spacing between blocks in 
Controller, Model and factories.

Still to do: 

Spacing on features, helpers and views.
Another run through on existing changes!


dotemacs:
Looks fine

PR #113, opened 2015-06-09
Added a constraint for User's Office

dotemacs:
Prevent the user from being created without being assigned to an
office. Added the constraint in the database schema.


colinbruce:
Looks good to me.

PR #114, opened 2015-06-09
Added a constraint for the User role

dotemacs:
There is a validation in place already, but just to be thorough, I've
added the constrain in the database schema.


colinbruce:
Good to merge :+1:

PR #116, opened 2015-06-10
Update markdown handler

colinbruce:
- remove erb string inject method
- replace with ruby object builder


dotemacs:
Go red team, go blue team!

PR #117, opened 2015-06-10
96296444 email changing

dotemacs:
Ticket: https://www.pivotaltracker.com/story/show/96296444

Disable editing of the emails for users by admins & managers.

All the other changes are in there because the foreign key was added to the **User** model on **office_id**.

Also, updated **RSpec** from 3.2.2 to 3.2.3.


colinbruce:
Looks good to me, merge at will.

PR #118, opened 2015-06-11
Handle undetermined results

colinbruce:
When a user searches using incorrect data
the system should return them to the input 
screen with a warning and allow them to 
amend the data.

[finishes #94226772]


dotemacs:
Good to go.

PR #121, opened 2015-06-17
95871446 dwp api healthcheck

dotemacs:
Added DWP API status to  **/healtcheck**

In the process I've tightened up the **HealthCheck** service.


colinbruce:
Looks ready to go...

PR #129, opened 2015-06-22
Link office and jurisdictions

colinbruce:
Create a link table between office and jurisdiction
* many-to-many link between the two models
* add jurisdiction check boxes to the office _form
* add jurisdiction display methods
* display the jurisdiction count on office index

Changes to existing objects
* a jurisdiction can be displayed as: name, name (abbr) and abbr only
* an office can have zero to many jurisdictions
* the office form views display jurisdiction checkboxes

### DIFF
--- a/spec/models/office_spec.rb
+++ b/spec/models/office_spec.rb
@@ -20,6 +20,35 @@ RSpec.describe Office, type: :model do
     it 'managers' do
       expect(office).to respond_to(:managers)
     end
+
+    it 'jurisdictions' do
+      expect(office).to respond_to(:jurisdictions)
+    end
+  end
+
+  describe 'jurisdiction' do
+    it 'can be nil' do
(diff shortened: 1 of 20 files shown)


spec/models/office_spec.rb
```diff
@@ -20,6 +20,35 @@
...
       expect(office).to respond_to(:managers)
     end
+
+    it 'jurisdictions' do
+      expect(office).to respond_to(:jurisdictions)
+    end
+  end
+
+  describe 'jurisdiction' do
+    it 'can be nil' do
```
dotemacs: Maybe here there can be:

```ruby
before(:each) { office.jurisdictions.clear }
```

instead of repeating 

```ruby
office.jurisdictions.clear
```

dotemacs:
Looks good.

PR #132, opened 2015-06-23
Link user to jurisdiction

colinbruce:
**Link user to jurisdiction**
Create a foreign key link from user to jurisdiction
* jurisdiction can be nil or a valid jurisdiction
* display the users' chosen jurisdiction on the index page
* display the users' chosen jurisdiction on the user/show view
* update the jurisdiction factory for invalid testing

Notes:
* No changes to the user/edit view as that will be done in 
the user profile ticket
* update the office tests to remove duplication of
the clear jurisdiction code

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,6 @@
 class User < ActiveRecord::Base
   belongs_to :office
+  belongs_to :jurisdiction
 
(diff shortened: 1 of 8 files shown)


dotemacs:
Good to go.

I'll merge it myself.

PR #137, opened 2015-06-29
Upgraded Rails to 4.2.3

dotemacs:
Don't need to lag behind and this release does not introduce any
regressions. Easy & simple upgrade.


dotemacs:
Retest this please

dotemacs:
./spec/models/office_spec.rb:80 failed previously. 

Ran in on my machine and it worked fine. It must be a race condition, that only manifests sporadically.

PR #139, opened 2015-06-30
Removal of R1 evidence

leighmoney:
Updated guide with info about evidence no longer being needed for R1


dotemacs:
@leighmoney looks good to me. I'll merge it now.

PR #147, opened 2015-07-03
Upgraded redcarpet gem due to vuln

dotemacs:
As per http://seclists.org/oss-sec/2015/q2/818
redcarpet has a vuln. This upgrade fixes it.


dotemacs:
I'm going to merge this as the failing test is addressed in #148 PR.

PR #163, opened 2015-07-07
Enforce all access to the app over SSL

dotemacs:
Use secure cookies.

This was one of the issues that was highlighted by the security
audit as done by @ohmygauss. So turning it on.


colinbruce:
Looks fine. Go for merge. :+1:

PR #164, opened 2015-07-07
Fix for user id in invitation url

colinbruce:
The user id should not be included in the invitation url.

It turns out that it isn't needed anyway.  Removing the 
resource and leaving the token is all that's required.

[finishes #94841546]


dotemacs:
Looks fine to me

PR #168, opened 2015-07-08
Static pages output to public

mattmb:
Fix mistake in output of static pages. Need to overwrite the pages in
/public.


colinbruce:
Code LGTM :+1:  You are go for merge :checkered_flag:

PR #170, opened 2015-07-08
Return 404 instead of 500

dotemacs:
When the user tries to look up some made up DWP check, it would return an error with 500 status code. Which means that the app would actually take the invalid input and do a database query with it. Which is not what we want.

This change prevents that by only allowing the valid format of the DWP check reference to be looked up.

### DIFF
--- a/app/controllers/dwp_checks_controller.rb
+++ b/app/controllers/dwp_checks_controller.rb
@@ -26,7 +26,7 @@ class DwpChecksController < ApplicationController
     authorize! :show, DwpCheck
   end
 
-private
+  private
 
   def new_from_params
     @dwp_checker = DwpCheck.new(dwp_params)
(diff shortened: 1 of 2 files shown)


dotemacs:
@mattmb in the response I'm returning 'public/404.html' page. But I remember that you've created a controller which would generate the error pages that are meant to be served from S3. It would be good if those error pages and the ones I'm referencing here (i.e. 'public/*.html' pages) are in sync.

mattmb:
I think you can just do: `render :status => 404` and rails will take care of the rest.

dotemacs:
@mattmb re-implemented the way the DWP checks were being looked up. So we don't have to worry about it, as it'll raise a exception which will serve up the 404 response.

mattmb:
LGTM :+1:

PR #171, opened 2015-07-08
Split assets generate and static page tasks

mattmb:
These tasks must be separate because we need rails to generate the
manifest for the assets before we create the static pages.


dotemacs:
Looks fine, merging

PR #174, opened 2015-07-08
Indent instead of outdent of keywords

dotemacs:
Changed the RuboCop setting so that the keywords are indented instead of
outdented. This applies to `private', `protected', ...


colinbruce:
Looks good :+1:

PR #175, opened 2015-07-09
Updating user jurisdiction management

colinbruce:
As discussed between @kelliematheson and @colinbruce.

Users will only be shown jurisdictions currently selected 
by office managers.

If managers have not chosen any jurisdictions, users are 
shown a warning message.


dotemacs:
LGTM

PR #177, opened 2015-07-09
Fixed markdown rendering bug

colinbruce:
Markdown was rendering, not executing, a script tag
the filter_html switch was the cause.

Removed the tag to ensure proper display

The JS was highlighting the page choice on the 
navigation bar.  This will be removed anyway once the 
design is updated.


dotemacs:
LGTM

PR #178, opened 2015-07-09
Update user style

colinbruce:
Update the styles of the edit user view

* Radio buttons should be used in preference to 
drop downs for small choices
* Field widths should be standardised across the
site


kelliematheson:
The style will require some further improvements still currently being defined, but this ticket is acceptable.

Happy for it to be merged.

PR #179, opened 2015-07-09
Updated manager/user editing

colinbruce:
Allows managers to update the office of a staff member

* It redirects to the list of their remaining users and displays
a flash message of where the user has transferred to.
* Replaced all references to User.first with explicit objects


dotemacs:
Small & clean, go

PR #181, opened 2015-07-09
Soft delete users

dotemacs:
As per subject: 'soft delete' users, but don't remove them from the database.


colinbruce:
LGTM. Go, go go! :checkered_flag:

PR #184, opened 2015-07-10
Allow users to change password

colinbruce:
Pivotal ticket 93656504

Awaiting content feedback before ready


stephenrichards:
Code reviewed: all OK

PR #185, opened 2015-07-13
Hide links from standard user

colinbruce:
Standard users should not see links to:
* delete their own account
* view the list of all users


dotemacs:
I'll merge.

PR #186, opened 2015-07-13
Prevent the manager from changing their role

dotemacs:
Don't allow a manager to change their role to user.

Added the examples for the scenario and removed the previous scenario
which did allow the manager to change his role.


colinbruce:
Looks good to me. :+1:

PR #190, opened 2015-07-20
updated gems

colinbruce:
* rspec-rails
* guard-rspec

after gemnasium warnings


colinbruce:
@mattmb Can you review this?

PR #191, opened 2015-07-20
Update table styles

colinbruce:
After the HR-fix from last week, there were some tables
that were not updated.

* fixed views
* updated style to match hide top that was used previously


colinbruce:
@mattmb style update

PR #203, opened 2015-07-28
The maintenance pages

dotemacs:
Updated maintenance pages with the copy as done by Leigh:

https://www.pivotaltracker.com/story/show/97121512

The pages 400 & 404 are the same, but leaving them as 'duplicate' just
in case that they might change in the future.


mattmb:
LGTM :+1:

PR #208, opened 2015-07-28
Privilege escalation prevention

dotemacs:
Privilege escalation prevention branch.

Workarounds for: 
* a user to prevent them from escalating their own privileges
* a manager to prevent them from escalating their user's or their own priviledges


colinbruce:
Looks good to me :+1:

PR #213, opened 2015-07-30
Point logs to STDOUT

mattmb:
Because we use docker in prod it makes sense to log to stdout and ship
from there.


mattmb:
build me

mattmb:
@colinbruce can you check this over

I've also added unicorn config to set the number of workers and some sane DB connection settings.

colinbruce:
@mattmb LGTM, merge at will! :checkered_flag:

PR #215, opened 2015-07-31
Refactored application params

colinbruce:
Moved the params into paged arrays and flattened
in the call, to allow better legibility of the list.


dotemacs:
Good to go

PR #221, opened 2015-07-31
Security update

colinbruce:
* updated ruby to 2.2.2
* updated sass-rails to 5.0.3

This will allow us to build against the updated docker templates


dotemacs:
Go!

PR #222, opened 2015-07-31
Savings summary

colinbruce:
Add the summary block to the foot of the 
savings and investment page.

This needed to access the applicant age 
from the model, so it was made public.


dotemacs:
Go!

PR #223, opened 2015-08-03
Update sign-in link css

colinbruce:
ensure consistent appearance with Sign out link


dotemacs:
I love the length of the PR change. I want more like this.

PR #229, opened 2015-08-03
Locale change

colinbruce:
Locale was set to en.

This meant that currency was formatting as $, etc.

Set locale to en-GB and update views to show currency in UK format.


dotemacs:
Go red team!

PR #230, opened 2015-08-04
Incorrect email error fix

colinbruce:
When an elevated user creates a new user, an enters an incorrect email
it was rendering a 500 error page.

This fix redirects them to the previous screen with the appropriate
error message.


dotemacs:
:+1:

PR #232, opened 2015-08-05
Style update users

colinbruce:
Update user views inline with new error styling.

These were created before the:
  Label
  Error
  Input

pattern was implemented


kelliematheson:
Approved, please merge

PR #234, opened 2015-08-07
Add benefit checks and views to application form

colinbruce:
When all of the required fields are completed and the 
application is saved, it will : 
* check that all fields are valid
* check that the fields are different from any previous check
and then fire off a request to the DWP.

This should be held until the income calculator is complete


stephenrichards:
This looks fine to me.

PR #235, opened 2015-08-07
Update copy on various pages

kelliematheson:
- Used term ‘log in’, now use ‘sign in’
- Used term ‘users’, now use ‘staff’
- Used term ‘edit’, now use ‘change’
- Used term ‘Invite’, now use ‘add’
- Used term ‘delete’, now use ‘remove’
Because it is more plain English


colinbruce:
LGTM Merge at will!

PR #240, opened 2015-08-10
Add view and tests for benefit results

colinbruce:


stephenrichards:
Yep, this looks fine too.

PR #242, opened 2015-08-11
Update copy on the office and user sections

kelliematheson:
Before said ‘edit’, now says ‘change’
Before said ‘default’, now says ‘main’
Before said ‘user’, now says ‘staff’


colinbruce:
Checked and passed :+1:

PR #250, opened 2015-08-17
Allow submission of 'empty' forms

colinbruce:
Added status field to views, this allows the submission of the form 
with no other fields completed.  This means that validations will fire
and return an error enforcing the completion of the question.


colinbruce:
@mattmb Can you review this?

mattmb:
LGTM, merge when the :traffic_light: goes green.

PR #253, opened 2015-08-20
Configure capybara javascript

colinbruce:
Updated rails_helper
* adds Database Cleaner configuration to allow javascript feature tests
to use capybara-webkit
* pre load support files

Update user tests  to remove explicit loading


colinbruce:
replaced by feature test branch

PR #268, opened 2015-09-08
Prevent the users from access '/dwp_check'

dotemacs:
The users should not be able to access '/dwp_check'.

I've left the examples by commenting them out: /it/(xit|pending)/,
because there will be a ticket in the backlog to deal with this
properly. This is just a band aid fix until tomorrow.


novotnyjakub:
I guess we don't want to merge this to master, this was just to build the branch?

PR #270, opened 2015-09-09
Using 'rake db:schema:load' for TravisCI

dotemacs:
Since Rails 4.1 this line in rails_helper:

    ActiveRecord::Migration.maintain_test_schema!

checks if there are any pending migrations and complains if there are.

Otherwise just load the schema.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ addons:
     repo_token: 9d2abe9e2bdfe62892cf37ae25ff3e825b07dad9401025adcb6f387ac82a35a8
 script:
 - bundle exec rake rubocop
-- bundle exec rake db:create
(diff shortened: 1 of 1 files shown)


dotemacs:
@mattmb What version of PostgreSQL are we using? I ask because in the TravisCI config we have **9.3** specified. Is there anything specific that needs to be pinned at this version or can we 'upgrade' the config to **9.4**? Thanks

mattmb:
9.3.5

We're limited to what RDS provides. It should auto update the minor version but won't bump to 9.4 until RDS release that version and we rebuild our stacks.

novotnyjakub:
RDS does provide `9.4`, we've used it for sure on other project.

novotnyjakub:
:+1:

PR #271, opened 2015-09-09
Run RuboCop first

dotemacs:
The idea is to have RuboCop run it's checks first because it's faster
than all the other tests. And if it is going to fail, it's better that
we know those failures straight away instead of waiting for all the
RSpec tests to pass and only then to be notified of the Ruby style
violations by RuboCop.


novotnyjakub:
:+1:

PR #278, opened 2015-09-14
Fix for r2 to r1 application forms

colinbruce:
See [Defect : Application form processing changing from R2 to R1 shows incorrect result](https://www.pivotaltracker.com/story/show/102991658)

[delivers #102991658]

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -132,6 +132,8 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
   def benefits=(val)
     super
     self.application_type = benefits? ? 'benefit' : 'income'
+    self.dependents = nil if benefits?
+    self.application_outcome = outcome_from_dwp_result if benefits? && last_benefit_check.present?
   end
 
   def full_name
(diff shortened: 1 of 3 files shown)


app/models/application.rb
```diff
@@ -132,6 +132,8 @@ def savings_investment_valid?
   def benefits=(val)
     super
     self.application_type = benefits? ? 'benefit' : 'income'
+    self.dependents = nil if benefits?
+    self.application_outcome = outcome_from_dwp_result if benefits? && last_benefit_check.present?
```
novotnyjakub: I'm thinking that it would be cleaner to have a method to convert the `Application` from `r1` to `r2`. Doing these assignments in setters is not obvious and I've ran into a similar problem in `children=` setter, which has a conditional in. 

I think setters should be setters and if there are conditions required for them, we should rather use validations or different methods. 

What does @dotemacs think?

colinbruce:
Turns out that this also resolves [Defect: Check details page showing wrong copy when switching savings and investments](https://www.pivotaltracker.com/story/show/102996704)

[finishes #102996704]

jitenp:
Passed testing.go go go

colinbruce:
Discussed with other dev and happy to push

PR #282, opened 2015-09-15
Next 16000

colinbruce:
'Next' button doesn't work on savings and investment page [Ticket](https://www.pivotaltracker.com/story/show/103322556)

---
Fix to allow progression past savings investments when applicant is over 61


dotemacs:
Yes, go ahead

PR #283, opened 2015-09-15
DWP stub trim

dotemacs:
Trimmed away at a stub request that is used frequently.

There are other instances of this stub being used, but they have more detailed responses configured. It might be an option to look at them and see if they need to be trimmed down.


colinbruce:
Nice, merge when ready.

PR #287, opened 2015-09-16
Fee threshold fix

colinbruce:
[Ticket:#103489954](https://www.pivotaltracker.com/story/show/103489954)

### DIFF
--- a/lib/fee_threshold.rb
+++ b/lib/fee_threshold.rb
@@ -1,0 +1,24 @@
+class FeeThreshold
+  def initialize(application)
+    @application = application
+  end
+
+  def band
+    return 3000 if fee <= 1000
+    return 4000 if fee.between?(1001, 1335)
+    return 5000 if fee.between?(1336, 1665)
+    return 6000 if fee.between?(1666, 2000)
+    return 7000 if fee.between?(2001, 2330)
+    return 8000 if fee.between?(2331, 4000)
+    return 10000 if fee.between?(4001, 5000)
+    return 12000 if fee.between?(5001, 6000)
+    return 14000 if fee.between?(6001, 7000)
+    return 16000 if fee >= 7001
+  end
+
+  private
+
+  def fee
+    @application.fee
+  end
+end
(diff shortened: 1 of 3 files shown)


colinbruce:
@novotnyjakub Can you code review this please?  It's a new, urgent, bug, but coded in a more more forward thinking way.

spec/lib/fee_threshold_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+RSpec.describe FeeThreshold do
+  let(:application) { create :application }
+  subject { described_class.new(application) }
+
+  context '.band' do
```
novotnyjakub: This is a really really minor thing, but `.` is by convention used for class methods and `#` is used for instance methods.

lib/fee_threshold.rb
```diff
@@ -0,0 +1,42 @@
+class FeeThreshold
```
novotnyjakub: I think this is really good preemptive extraction for the refactoring we're doing :+1:

lib/fee_threshold.rb
```diff
@@ -0,0 +1,42 @@
...
+      { lower: 1336, upper: 1665, amount: 5000 },
+      { lower: 1666, upper: 2000, amount: 6000 },
+      { lower: 2001, upper: 2330, amount: 7000 },
+      { lower: 2331, upper: 4000, amount: 8000 },
+      { lower: 4001, upper: 5000, amount: 10000 },
+      { lower: 5001, upper: 6000, amount: 12000 },
+      { lower: 6001, upper: 7000, amount: 14000 }
+    ]
+
+  def initialize(application)
```
novotnyjakub: I'd maybe pass in the fee, rather than the whole application, so that it'd be less coupled with the `Application` model?

dotemacs: @novotnyjakub This change is not a refactor per se, but it's rather a urgent bug fix that needs to go in. This was the bug that Jiten tried to talk to you about this morning.

PR #288, opened 2015-09-16
Handle decimal fees

dotemacs:
Added the ability to work out the fee threshold band even when the fee
is supplied as a decimal value. Previously it only handled whole
numbers.


dotemacs:
I'll re-raise this PR against the master branch.

Closing for now...

PR #289, opened 2015-09-16
Fee threshold rounding

dotemacs:
Added the ability to work out the fee threshold band even when the fee
is supplied as a decimal value. Previously it only handled whole
numbers.

Also, passing in just the fee instead of the whole Application model.

### DIFF
--- a/lib/fee_threshold.rb
+++ b/lib/fee_threshold.rb
@@ -28,7 +28,7 @@ class FeeThreshold
   private
 
   def fee
-    @application.fee
+    @application.fee.round
   end
 
   def find_band(line)
--- a/lib/fee_threshold.rb
+++ b/lib/fee_threshold.rb
@@ -11,14 +11,15 @@ class FeeThreshold
       { lower: 6001, upper: 7000, amount: 14000 }
     ]
 
-  def initialize(application)
-    @application = application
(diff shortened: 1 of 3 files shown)


lib/fee_threshold.rb
```diff
@@ -11,14 +11,15 @@ class FeeThreshold
...
+  def initialize(fee)
+    return if fee.nil?
+    @fee = fee.round
   end
 
   def band
-    return if fee.nil?
-    return 3000 if fee <= 1000
-    return 16000 if fee >= 7001
+    return if @fee.nil?
```
novotnyjakub: don't need this here as `fee` can't be `nil` (prevented in the initializer).

novotnyjakub:
Looks good.

lib/fee_threshold.rb
dotemacs: Writing this for historical reasons, as we've already discussed this in person.

By removing the nil guard in the `band` method, I would get bitten, see:

```
f = FeeThreshold.new(nil)
=> #<FeeThreshold:0x007fa469a387b0>
[4] pry(main)> f.band
=> nil
[5] pry(main)> f.class
=> FeeThreshold
```

PR #292, opened 2015-09-18
Add sentry to capture exceptions.

mattmb:
Also add a raise exception method to test the capture.


mattmb:
@colinbruce @dotemacs simple one to review. Let me know if you'd rather pull out the raise_exception method, seems useful for testing but happy to pull it out if you want to keep it clean.

dotemacs:
@mattmb & me had a chat and I suggested that the method for raising the exception be moved to **health_status_controller.rb** as that is where all the health checking fun happens

colinbruce:
Looks good... Merge

PR #293, opened 2015-09-18
Refactor to form objects

dotemacs:
It started of as the first step towards refactoring all the pages so that they use form objects.

But we've added the refactoring of the **ApplicationDetail** which holds the second page **application_details**. 

And we decided to use this branch as the home for all the subsequent changes.

:octocat: :syringe: :eyes:


colinbruce:
Looks good :+1:

colinbruce:
Looks better! :checkered_flag:

dotemacs:
@novotnyjakub can you please have a look at this when you get a minute? Thanks

dotemacs:
@colinbruce oh, and there is only **one** failing test :bangbang:

dotemacs:
@colinbruce 

I think that this is the problematic part:
https://github.com/ministryofjustice/fr-staffapp/blob/personal-details-refactor/app/controllers/applications/build_controller.rb#L36-L37

The controller should redirect after `@application.savings_investment_valid?` but looking at **savings_investment_valid?**: 
https://github.com/ministryofjustice/fr-staffapp/blob/personal-details-refactor/app/models/application.rb#L132-L139 it seems to me that this is where we have to have a better look.

colinbruce:
@dotemacs This can be squashed too at some point, but the setting on the page was wrong.

The tests, and validation, now pass :question: :exclamation:

dotemacs:
@colinbruce oh, the reason for failures is RuboCop violations! Great:)

colinbruce:
Looks good and tests passing... go, go, go.

PR #295, opened 2015-09-22
Updated admin dashboard

colinbruce:
Following the release of the application form, the checks are recorded differently

This allows us to track the new benefit checks and display the admin 
graphs for new dwp checks.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,10 +1,14 @@
 class HomeController < ApplicationController
   before_action :authenticate_user!, only: [:index]
-  before_action :load_graph_data, only: [:index], if: 'user_signed_in? && current_user.admin?'
 
   def index
     manager_setup = ManagerSetup.new(current_user, session)
     manager_setup.finish! if manager_setup.in_progress?
+    if current_user.admin?
+      load_graph_data
+      @total_type_count = BenefitCheck.group(:dwp_result).count
+      @time_of_day_count = BenefitCheck.group_by_hour_of_day("created_at", format: '%l %p').count
+    end
   end
 
   private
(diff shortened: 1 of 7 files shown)


db/migrate/20150921084620_add_timestamps_to_benefit_check.rb
```diff
@@ -0,0 +1,19 @@
+class AddTimestampsToBenefitCheck < ActiveRecord::Migration
+  def up
+    change_table(:benefit_checks) { |t| t.timestamps }
+    migrate_sql = <<SQL
```
dotemacs: My comment here is to look at this way of formatting: https://github.com/bbatsov/ruby-style-guide#heredocs

dotemacs:
Apart from the comment above, it looks fine.

app/controllers/home_controller.rb
```diff
@@ -1,6 +1,5 @@
 class HomeController < ApplicationController
   before_action :authenticate_user!, only: [:index]
-  before_action :load_dwp_data, only: [:index], if: 'user_signed_in? && current_user.manager?'
   before_action :load_graph_data, only: [:index], if: 'user_signed_in? && current_user.admin?'
```
novotnyjakub: Isn't it cleaner to call `load_graph_data` directly in the `index` action, rather then using filters? Especially if the filter has a complex condition and is used only in a single action.

app/views/home/_admin.html.slim
```diff
@@ -1,11 +1,11 @@
 .row.collapse
   .columns.small-12
     h2 Total
-    = column_chart DwpCheck.non_digital.group(:dwp_result).count, { library: { title: 'All time responses by type', width: '200px' } }
+    = column_chart BenefitCheck.group(:dwp_result).count, { library: { title: 'All time responses by type', width: '200px' } }
```
novotnyjakub: I think we should avoid doing queries in the views?

dotemacs:
Looks fine

PR #298, opened 2015-09-28
Fix for rubocop blocking vagrant provision

colinbruce:
When the vagrant box was started locally, the rubocop rake task 
would block __any__rake task.  This prevented rake db:migrate
running.

This fix prevents rubocop running when DOCKER_STATE = 'vagrant', 
but still runs if rake is used locally.


dotemacs:
Looks fine

PR #299, opened 2015-09-28
Making vagrant work

colinbruce:
Local vagrant deploy was working, but the database was not being seeded.

To fix this: 

* I updated the seed file to create offices with pre-selected jurisdictions, this satisfies the
new validation created by the out-of-box experience feature
* Made sure that the run.sh file runs migrate _before_ attempting to seed the data


mattmb:
Looks very sane to me :+1: 

We don't need the offices in the real world to have that data do we. i.e. there's no need to write a migration?

colinbruce:
@mattmb Correct.  This is purely for dev/test environments.

PR #300, opened 2015-09-29
Prevent user deleting themselves

colinbruce:
Update user handling to prevent deleting themselves
* hide the link from user if viewing own record
* route them back to the details page with an error if
they spoof the delete function

[finishes #103483020]

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,9 +2,7 @@ class UsersController < ApplicationController
   respond_to :html
   before_action :authenticate_user!
   before_action :find_user, only: [:edit, :show, :update, :destroy]
-  before_action :populate_roles, only: [:edit, :update]
-  before_action :populate_offices, only: [:edit, :update]
-  before_action :populate_jurisdictions, only: [:edit, :update]
+  before_action :populate_lookups, only: [:edit, :update]
   load_and_authorize_resource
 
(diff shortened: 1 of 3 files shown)


GitCop:

 * Commit: 09cdf837b58b9b61b4ee6d43955f1fefd3d8570e

---

app/controllers/users_controller.rb
```diff
@@ -2,9 +2,7 @@ class UsersController < ApplicationController
   respond_to :html
   before_action :authenticate_user!
   before_action :find_user, only: [:edit, :show, :update, :destroy]
```
dotemacs: I know that you didn't touch this line, but what do you think of this:

`before_action :find_user, except: [:index]`

?

app/views/users/show.html.slim
```diff
@@ -33,10 +33,11 @@ h2 Staff details
   .row.pad-top-fifteen-px
     .columns.small-12
       = link_to "Change your password", edit_user_registration_path(@user)
--if current_user.elevated?
+-if current_user.elevated? && current_user != @user
```
dotemacs: Here I'd put some sort of method that would describe the condition in an clear way. Something like:

`if user_has_the_permissions_and_not_themselves`

dotemacs:
Discussed it offline, approved

PR #301, opened 2015-09-29
Eased out benefits page to a form object

dotemacs:
Created Benefit model which only handles the benefit page form. This is
done as part of the refactoring task on the monolithic Application
model.

Also removed the 'status' attribute from the benefits page and addded a
hidden field, this is because otherwise the form would not submit any of
the fields from the radio buttons.


colinbruce:
Looks fine.

novotnyjakub:
:+1:

PR #302, opened 2015-09-29
Income refactor

dotemacs:
Refactor of **income** page into a form object, in order to get away from the monolithic **Application** model.


colinbruce:
LGTM

PR #303, opened 2015-09-30
Admin user delete

colinbruce:
There was a need for admins to undelete users.

This adds a new users/deleted view that only lists deleted users
The table has a per user restore function

As part of the work, I refactored some of the complexity to a new UserManagement 
service class.  This is called from the UserController and the show view to determine
eligibility to delete.


dotemacs:
Very cool! Nice PR.

PR #304, opened 2015-09-30
Add Email link to feedback display

colinbruce:
This will allow admins to contact users directly
without having to search for users and copy/paste
email addresses from the user table.


dotemacs:
Approved, go

PR #305, opened 2015-10-01
Update static page rake task

colinbruce:
The static pages were being generated with www.example.com because
(as they are static!) they did not have access to root_url.

This replaces http://www.example.com with /, thus taking users
to the root page of the current site, regardless of environment.


dotemacs:
Good, do it

PR #306, opened 2015-10-01
DOB validations

dotemacs:
When refactoring the monolithic Application model to be extracted to
Personal Informaton form object, the validation for the minimum/maximum age was
not brought over.


colinbruce:
Looks fine

PR #307, opened 2015-10-01
I18n cleanup in User model

dotemacs:
The model shouldn't know the full path to the locale error messages,
they should be derived upon validation.


colinbruce:
Go! :checkered_flag:

PR #308, opened 2015-10-01
Stub out the requests before each context

dotemacs:
From my reading of:

https://www.relishapp.com/rspec/rspec-core/v/3-2/docs/hooks/before-and-after-hooks

it seems that having the stubs set before each context, fixes the
intermitten failures.


dotemacs:
@colinbruce please have a read of the above URL I posted and tell me what you think. Thanks

PR #310, opened 2015-10-02
Updated arrow display for firefox

colinbruce:
Updated style to reference a span inside the summary
updated the views to have the required span


dotemacs:
I believe you

PR #311, opened 2015-10-02
Remove js from view

dotemacs:
Removed JavaScript from the view on Savings and Investments page.

Just cleaning up...


colinbruce:
Yes, looks fine.

PR #313, opened 2015-10-05
Refactor spot to evidence check

colinbruce:
Due to a change of policy, spot checks are now known as evidence checks.

To prevent inconsistencies between the code base, user knowledge and experience;
all references to spotchecks are now known as evidence_checks.


dotemacs:
Seeing as this is a straightforward rename, no issues here, feel free to merge :+1:

PR #314, opened 2015-10-05
Introduced a new namespace under models

dotemacs:
Since there will be two major flows:

1. for Application processing
2. for Evidence processing

there will be different form objects that handle their forms. So
introducing namespacing in order to organise the code, so that anybody
looking at the codebase can know where the models for Application sit
and where models for Evidence sit.

Used the name 'Applikation' as it was conflicting with 'Application' and
Rails server would not even start.


app/models/form_object.rb
```diff
@@ -0,0 +1,29 @@
+class FormObject
```
dotemacs: Added this so that it can be shared by both namespaces **Applikation** & **Evidence** (when it comes)

colinbruce:
Looks good

PR #315, opened 2015-10-05
Set feedback to retain deleted users details

colinbruce:
When feedback was displayed, if a deleted user had left 
feedback the page would error.  This sets the association to 
include with_deleted to prevent the error.


dotemacs:
Thanks for taking care of this.

Should there be a test that catches the error first?

colinbruce:
Noted: I'll create the tests and re-push

PR #317, opened 2015-10-05
Created Evidence form object

dotemacs:
Created a model which will handle the evidence form values and their
validation.


colinbruce:
Looks good.

PR #318, opened 2015-10-05
Fix for migration error

colinbruce:
Previous migration script ran a per object script
A the object was renamed, it could not resolve the object to
the database. I switched it to a sql script to run purely
agains the table and it now works on local machines.

Lets see how branch builder does...


dotemacs:
If you've checked the instance on docker and it seeds the database as it should, then great, go ahead

colinbruce:
:+1:

PR #319, opened 2015-10-05
Remove rubocop exclusion

colinbruce:
This is not needed after a renaming of the object and
was generating a reubocop warning


dotemacs:
:+1:

PR #320, opened 2015-10-05
Evidence income form view

dotemacs:
Despite what the branch name says, this is the Income form object.
It's to be used to validate the Income amount the user provides when assessing the evidence based check.

### DIFF
--- a/app/models/evidence/forms/income.rb
+++ b/app/models/evidence/forms/income.rb
@@ -1,0 +1,16 @@
+module Evidence
+  module Forms
+    class Income < ::FormObject
+
+      def self.permitted_attributes
+        {
+          income: String
+        }
+      end
+
+      define_attributes
+
+      validates :income, presence: true
+    end
+  end
+end
--- a/app/models/evidence/forms/income.rb
+++ b/app/models/evidence/forms/income.rb
@@ -4,14 +4,14 @@ module Evidence
 
       def self.permitted_attributes
         {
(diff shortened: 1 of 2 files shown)


dotemacs:
@colinbruce please have a look at this when you get a minute, thanks

app/models/evidence/forms/income.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      def self.permitted_attributes
+        {
+          amount: String
+        }
+      end
+
+      define_attributes
+
+      before_validation :format_amount
```
dotemacs: I'm using this callback now, but when it comes to persisting this data, I'll remove the callback and I'll just reuse the **format_amount** method to do the rounding of the amount before saving it.

Something like this:

```ruby
def save
   if valid?
     persist!
     true
   else
     false
   end
 end

private

def persist!
  @evidence = Evidence.create!(income: format_amount(amount))
end
```

novotnyjakub:
@dotemacs is this PR active, or should we close it for now? It's to be merged to `master`, so that's why I'm asking.

dotemacs:
I'll be merging this into **evidence-controller** branch as I'll be using the form object implemented here for the page 4. of the prototype.

novotnyjakub:
:+1:

dotemacs:
Closing and re-opening PR on another branch

PR #321, opened 2015-10-06
We're not going to use Concerns

dotemacs:
And we haven't so far, so removing it.


colinbruce:
Looks good, merge at will.

PR #323, opened 2015-10-06
Proposed method of styling the overview

colinbruce:
@dotemacs This is what I was trying to explain.

Feel free to delete if you think it is unworkable!

I created the tests and methods for the _css overloads, and 
called them from the view.

I refactored the hashes to reduce the duplication


colinbruce:
Superseded by different PR

PR #324, opened 2015-10-07
Evidence controller

dotemacs:
**DO NOT MERGE JUST YET!**

Evidence check epic.

Currently it has the following implemented:

- page 1.
- page 2.
- page 3.
- page 4.
- page 5.
- page 6.
- page 7.


GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

GitCop:

 * Commit: 4ec62a338077d08c4cb98cfee0c60fbd2253be38

---

app/models/reason.rb
```diff
@@ -0,0 +1,3 @@
+class Reason < ActiveRecord::Base
```
novotnyjakub: I seriously think we should have this model. `Reason` is very generic and will be very hard to follow in the future when looking at the codebase. Also it means extra SQL queries when working with `EvidenceCheck`. 

If you insist on having a dedicated table for this 1 field, let's at least rename it to something like `EvidenceCheckInvalidReason` or something between those lines, so it's clear what this is for.

dotemacs: I'm guessing your first sentence here was going to be:

    I seriously think we should NOT have this model. 

If that's the case... I don't think that having to write a query to fetch Reasons for EvidenceChecks having failed:

- because this app will not have that many users that millisecond optimisations are going to hold us back
- it never hurts to have the data organised in separate components
- in the future the reason text field will potentially be replaced with a radio box, by having it in a separate table it's easier to extend/add that functionality 
- as for the CamelCasedName of **EvidenceCheckInvalidReason** I think that it looks ugly. The 'belongs_to' clearly describes what the table relates to and I find this easier:

```ruby
evidence.reason.explanation
```
than
```ruby
evidence.evidencecheckinvalidreason.explanation
```

novotnyjakub: OK, let's catch up about this tomorrow, ideally with @colinbruce as well. 
Future proofing shouldn't be our concern at the moment and right now this will make things more complicated. For example now can create multiple reasons for 1 application, which is a case we'll have to handle as well. Also the user will have to edit the existing reason, which means the form has to pull data from 2 tables.

dotemacs: @colinbruce is off today. And we already discussed it in person yesterday.

We're not future proofing this as it's difficult to predict the future. What I believe that we're doing here, is that we're creating a model (in the modelling sense, not ActiveRecord's use of model) of something that can be easily extended if need be.

As for:

> For example now can create multiple reasons for 1 application, which is a case we'll have to handle as well.

I haven't see this anywhere, happy to be proved wrong.
With the current setup of the Evidence form object the user should only ever have one reason. If they choose to add another reason, they can just append it to the existing **explanation** attribute.

> Also the user will have to edit the existing reason, which means the form has to pull data from 2 tables.

And that is bad because?

colinbruce:
As agreed with @novotnyjakub and @dotemacs, everything is green, merging into master...

PR #328, opened 2015-10-07
Styling evidence summary pages

colinbruce:
Added css styling to the new summary page to match
the design.

Checks for ✓ or ✗ in the value and applies the current 
result (full, part or none) to the value row.

The styles are updated to ensure that the names match
the (current) outcomes.

### DIFF
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -24,11 +24,18 @@ module SummaryHelper
     label = I18n.t("activemodel.attributes.#{object.class.name.underscore}.#{field}")
     value = object.send(field)
 
+    value_classes = 'small-12 medium-7 large-8 columns'
+    value_classes << " summary-result #{object.result}" if value_needs_styling?(value)
+
     rows = content_tag(:div, label, class: 'small-12 medium-5 large-4 columns subheader')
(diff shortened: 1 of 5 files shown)


app/helpers/summary_helper.rb
```diff
@@ -24,11 +24,18 @@ def build_data_row(object, field)
...
-    rows << content_tag(:div, "#{value}".html_safe, class: 'small-12 medium-7 large-8 columns')
+    rows << content_tag(:div, value, class: value_classes)
 
     content_tag(:div, class: 'row') do
       rows
     end
   end
+
+  def value_needs_styling?(value)
+    %w[✓ ✗].any? { |char| value.to_s.include? char }
```
novotnyjakub: Interesting way of doing it :)

config/locales/en.yml
```diff
@@ -52,9 +52,9 @@ en-GB:
       wrong_office: You are not authorized to manage this user.
   benefit_checks:
     :yes:
-      heading: '&#10003; &nbsp; The applicant is receiving the correct benefits'
+      heading: '✓ &nbsp; The applicant is receiving the correct benefits'
```
novotnyjakub: I think we should consider adding the tick and cross using CSS, but that's not important now.

novotnyjakub:
Waiting for the build to pass and then I'll merge.

PR #329, opened 2015-10-08
Updated accuracy page

colinbruce:
Moved to locale file for labels
Switched input to text-area
Updated foundation use for optional panel

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -156,6 +156,9 @@ en-GB:
         <p>%{user_name}</p>
   activemodel:
     attributes:
+      evidence/forms/evidence:
+        correct: Is the evidence correct?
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -156,6 +156,9 @@ en-GB:
         <p>%{user_name}</p>
   activemodel:
     attributes:
+      evidence/forms/evidence:
+        correct: Is the evidence correct?
+        reason: What is incorrect about the evidence?
```
novotnyjakub: There's an extra article here, which breaks the build :)

novotnyjakub:
If you don't mind fixing the test, then I'm happy to merge :+1:

colinbruce:
Cheers @novotnyjakub
**The** tests are re-running now!

PR #332, opened 2015-10-08
Implement evidence calculation result

colinbruce:
Added a Result view model, it returns the EvidenceCheck.outcome 
and amount to pay.

This is rendered on the view using the remission partial.

The EvidenceCheck factory was updated to include the outcome
and amount_to_pay fields.


dotemacs:
Not sure about the xpath checking of the content... but apart from that it looks fine.

PR #333, opened 2015-10-08
Waiting for evidence dashboard

novotnyjakub:
It adds the list of applications waiting for evidence to the dashboard for managers and staff. 

I'm not particularly happy with some bits of the implementation, but I tried to re-use as much as possible from what we already have, so we can move on to the next bits of the evidence check flow.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -12,6 +12,10 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
     where(benefits: false, application_type: 'income', application_outcome: %w[part full])
   }
 
+  scope :waiting_for_evidence, lambda {
+    includes(:evidence_check).references(:evidence_check).where.not(evidence_checks: { id: nil })
+  }
+
   MAX_AGE = 120
   MIN_AGE = 16
 
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -199,6 +199,18 @@ RSpec.describe Application, type: :model do
         is_expected.to match_array([application_1, application_2])
       end
     end
+
+    describe '.waiting_for_evidence', focus: true do
+      let!(:application1) { create :application }
+      let!(:application2) { create :application }
+      let!(:evidence_check) { create :evidence_check, application: application1 }
+
+      subject { described_class.waiting_for_evidence }
+
+      it 'returns only applications which have EvidenceCheck reference' do
+        is_expected.to match_array([application1])
+      end
+    end
   end
 
   describe 'ni_number' do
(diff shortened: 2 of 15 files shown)


colinbruce:
Looks good :+1: , apart from the rubocop complaint! :-1:

app/models/application.rb
```diff
@@ -12,6 +12,10 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
     where(benefits: false, application_type: 'income', application_outcome: %w[part full])
   }
 
+  scope :waiting_for_evidence, lambda {
+    includes(:evidence_check).references(:evidence_check).where.not(evidence_checks: { id: nil })
+  }
+
```
dotemacs: I'd avoid adding anything else to the Application model, as it's fat enough as is.

So I'd rather we have some sort of Query model, that just constructs this query. As per point 4. in:
http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/

spec/models/application_spec.rb
```diff
@@ -199,6 +199,18 @@
         is_expected.to match_array([application_1, application_2])
       end
     end
+
+    describe '.waiting_for_evidence', focus: true do
+      let!(:application1) { create :application }
+      let!(:application2) { create :application }
+      let!(:evidence_check) { create :evidence_check, application: application1 }
```
dotemacs: Can we use mocks/stubs here?

As it's a model and this is meant to be a unit test.

novotnyjakub:
Yeah, I kept an eye on it, but the last commit :) Fixing now..

spec/models/application_spec.rb
novotnyjakub: This is an AR scope and that hits the DB, so this is the only way of testing it.

app/models/application.rb
novotnyjakub: I agree that the model is fat at the moment (work in progress), but scopes are the things that belong here, imho. Lot of the other stuff doesn't. 

I'm not against moving this to a query model, however I'd wait until we finish the evidence check flow and then know what these queries actually look like. I wasn't even sure if this is a concern of `Application` or `EvidenceCheck`. But `EvidenceCheck` does not belong to `Office`, therefore I chose to put it here.

dotemacs: Just had a chat with @colinbruce and we both agree to go with this **for now**, just to complete Evidence check flow, but we must refactor Application model, so I hope that this doesn't get left behind, especially since we've got another epic after this one is done!

colinbruce:
Happy for this to merge, with the final caveat above :) 
:checkered_flag: go, go, go!

PR #334, opened 2015-10-08
Remove form column from dashboard list

novotnyjakub:


dotemacs:
I can see what you've done from the diff & the git subject, but why was it done?

novotnyjakub:
I asked Kellie, if we should really render this field on the dashboard and she said we shouldn't. So I removed it.

PR #335, opened 2015-10-08
Evidence summary page

novotnyjakub:
It's ready to be reviewed / merged now. 

It brings the "page No. 6" part of the evidence flow. When the evidence is not correct, the page is redirected straight to this summary page. When the evidence is correct, the summary page is displayed after income result page.

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -23,6 +23,12 @@ class EvidenceController < ApplicationController
     evidence_result
   end
 
+  def summary
+    evidence_view
+    evidence_overview
+    evidence_result
+  end
+
   private
 
   def prepare_evidence
@@ -32,6 +38,11 @@ class EvidenceController < ApplicationController
   def evidence_overview
     prepare_evidence
     @overview = Evidence::Views::Overview.new(@evidence)
+  end
+
+  def evidence_view
+    prepare_evidence
+    @evidence_view = Evidence::Views::Evidence.new(@evidence)
   end
 
   def accuracy_form
@@ -46,9 +57,17 @@ class EvidenceController < ApplicationController
     @form = Evidence::Forms::Evidence.new(evidence_params)
 
     if @form.save
-      redirect_to evidence_income_path
+      redirect_after_accuracy_save
     else
       render :accuracy
+    end
+  end
+
+  def redirect_after_accuracy_save
+    if @form.correct
+      redirect_to evidence_income_path
+    else
+      redirect_to evidence_summary_path
     end
   end
 
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -35,16 +35,12 @@ module SummaryHelper
   end
 
   def value_style(value)
-    base_class = 'small-12 medium-7 large-8 columns'
-    extended_class = case value.to_s
-                     when /✓/
-                       ' summary-result passed'
-                     when /✗/
-                       ' summary-result failed'
-                     else
-                       ''
-                     end
-
-    base_class + extended_class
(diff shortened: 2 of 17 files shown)


app/helpers/summary_helper.rb
```diff
@@ -24,18 +24,27 @@ def build_data_row(object, field)
...
+  def value_style(value)
+    base_class = 'small-12 medium-7 large-8 columns'
+    extended_class = case value.to_s
+                     when /✓/
+                       ' summary-result passed'
+                     when /✗/
+                       ' summary-result failed'
+                     else
+                       ''
+                     end
```
dotemacs: A more succinct version of that case statement can be implemented in the following way:

```ruby
def value_style(value)
  ['small-12 medium-7 large-8 columns',
   ({ '✓' => ' summary-result passed',
      '✗' => 'summary-result failed' }[value.to_s] || '')].join
end
```

config/routes.rb
```diff
@@ -11,6 +11,7 @@
   get 'evidence/:id/income', to: 'evidence#income', as: :evidence_income
   post 'evidence/:id/income_save', to: 'evidence#income_save', as: :evidence_income_save
   get 'evidence/:id/result', to: 'evidence#result', as: :evidence_result
+  get 'evidence/:id/summary', to: 'evidence#summary', as: :evidence_summary
```
dotemacs: What checks this route?

I ask only because all the other routes for this controller have the route specs.

spec/factories/reasons.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :reason do
+    explanation 'EXPLANATION'
+  end
+end
```
dotemacs: Where is this factory used?

spec/models/evidence/forms/evidence_spec.rb
```diff
@@ -60,17 +60,52 @@
...
   end
 
-  describe '#save' do
-    before do
-      allow(subject).to receive(:valid?).and_return(true)
-      allow(subject).to receive(:persist!)
-      allow(EvidenceCheck).to receive(:find)
-      allow_any_instance_of(EvidenceCheck).to receive(:update)
-      allow_any_instance_of(Reason).to receive(:save)
+  describe '#save', focus: true do
```
novotnyjakub: I've changed the testing approach to this method, even though I know it'll create a discussion around it. So I'd like to explain:

Because it's now setting outcome of the whole evidence flow, when the evidence is incorrect, I feel that this needs to be tested solid and making sure the right fields are updated on the right columns. 
So it's a bit slower than it was, but I think it's more robust. If we refactored the form to accept the `EvidenceCheck` in the constructor, we wouldn't have to store the evidence check in the database in the test. I think it would be cleaner to test.

dotemacs: By taking this approach, you've turned the unit test into an integration test and that can be done with the features rather than testing it here.

I'm not going to oppose this, as I'd rather see the feature complete. But I'd like to revisit this and try to implement it in a cleaner way without touching the database.

app/controllers/evidence_controller.rb
```diff
@@ -23,6 +23,12 @@ def result
     evidence_result
   end
 
+  def summary
+    evidence_view
+    evidence_overview
+    evidence_result
+  end
```
dotemacs: This begs to be cleaned up.

I've got to 'jump' up and down to figure out why and what is going on. The intent is not clear.

dotemacs:
@colinbruce we've discussed this offline so to keep you in the loop: we'll merge this just to get the whole flow going, but we'll need to re-do the controller as it's a bit messy.

PR #337, opened 2015-10-09
Evidence result specs

dotemacs:
Added the missing specs to the result action on the evidence controller.

All the other actions have these specs, so trying to stay uniform.

### DIFF
--- a/spec/controllers/evidence_controller_spec.rb
+++ b/spec/controllers/evidence_controller_spec.rb
@@ -103,4 +103,19 @@ RSpec.describe EvidenceController, type: :controller do
       end
     end
   end
+
+  describe 'GET #result' do
+    before do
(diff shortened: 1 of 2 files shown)


spec/controllers/evidence_controller_spec.rb
```diff
@@ -103,4 +103,19 @@
       end
     end
   end
+
+  describe 'GET #result' do
```
novotnyjakub: Do we need this, if we have feature tests covering the same thing? If we were testing some assignments here, maybe, but we only test response and template, whereas in feature and view model specs we test all aspects of this page.

GitCop:

 * Commit: f96e4b4f6174991a14d58573a0f1a1f760c7115f

---

colinbruce:
Looks fine... :checkered_flag:

PR #338, opened 2015-10-09
Evidence confirmation

dotemacs:
Still not polished, as it doesn't have the view object for passing in the data for the views cleanly.

But apart from that, mostly done.

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -24,6 +24,7 @@ class EvidenceController < ApplicationController
   end
 
   def confirmation
+    evidence_confirmation
   end
 
   private
@@ -75,6 +76,10 @@ class EvidenceController < ApplicationController
   def evidence_result
     prepare_evidence
     @result = Evidence::Views::Result.new(@evidence)
+  end
+
+  def evidence_confirmation
+    @confirmation = EvidenceCheck.find(params[:id])
   end
 
   # TODO: permitted params setup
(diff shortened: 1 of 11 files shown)


app/controllers/evidence_controller.rb
```diff
@@ -74,5 +78,9 @@ def evidence_result
     @result = Evidence::Views::Result.new(@evidence)
   end
 
+  def evidence_confirmation
+    @confirmation = EvidenceCheck.find(params[:id])
```
novotnyjakub: Just a little thing, I'd call this `@evidence_check` rather than `@confirmation`, so it's clear in the template what the object represents.

spec/features/evidence/evidence_check_flow_spec.rb
```diff
@@ -101,4 +101,30 @@
...
+  context 'when on "Evidence confirmation" page' do
+    before { visit evidence_confirmation_path(id: evidence.id) }
+
+    it { expect(page).to have_content 'Processing complete' }
+
+    context 'when the remission is' do
+      context 'full' do
+        let(:outcome) { 'full' }
+
+        it { expect(page).not_to have_content /(not\ correct\|part-fee)/ }
```
novotnyjakub: Capybara has a special matcher for testing that something is not present on the page: `have_no_content()`, because using `not_to` doesn't take into account Javascript wait, etc.

But I think that this test may not be relevant anyway, because testing that a particular text is not on the page doesn't proof that the page is correctly rendered. Hope that makes sense.

PR #339, opened 2015-10-09
Removed unused DWP

dotemacs:
Removed all the tests and the classes in the app, as we're no longer
using this feature.

The main reason for doing this is that it just sat around and made the
output of the tests cluttered. Also there's a ticket in the backlog to
remove this...


dotemacs:
@colinbruce can we talk about [this line](https://github.com/ministryofjustice/fr-staffapp/blob/remove-unused-dwp/spec/services/partner_age_check_spec.rb#L19) ?

novotnyjakub:
This is much appreciated. I would just suggest to merge this to the `evidence-controller` branch, rather than `master`? Considering that for the moment that branch is kind of our master. I'd also prefer to rebase on these changes.

dotemacs:
I'll follow @novotnyjakub advice and I'll close this PR as I pointed it at **master** and instead I'll point it at **evidence-controller** branch.

PR #340, opened 2015-10-09
Allow to change evidence

novotnyjakub:
The purpose of this PR is to allow going back from the summary page and changing the evidence if necessary. 

I haven't finished all the refactoring I wanted, but it's not making things worse. I think we can merge this now, so that the evidence is editable and then further improve the `EvidenceController`.

### DIFF
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -22,7 +22,8 @@ module SummaryHelper
 
   def build_link(link_title, link_url)
     if link_title.present? && link_url.present?
-      content_tag(:div, class: 'small-12 medium-4 large-3 columns') do
+      link_class = 'small-12 medium-5 large-4 columns medium-text-right large-text-right'
+      content_tag(:div, class: link_class) do
         link_to(link_title, link_url)
       end
     end
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -1,8 +1,8 @@
 module SummaryHelper
 
-  def build_section(summary_text, object, fields)
+  def build_section(summary_text, object, fields, link_title = nil, link_url = nil)
     content_tag(:div, class: 'summary-section') do
-      content = build_header summary_text
+      content = build_header(summary_text, link_title, link_url)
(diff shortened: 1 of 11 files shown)


novotnyjakub:
ok @dotemacs so this is almost done, apart from the parameters handling in the `EvidenceController`. The problem is that the params come from 3 different sources (url, `EvidenceCheck` model, `Reason` model), so there needs to be logic in the controller to handle that. 

What I'm gonna do is try to refactor the evidence forms to what I think could solve this problem and push it up. You most probably won't like it :) but I'll give it a shot.

novotnyjakub:
@dotemacs @colinbruce I haven't finished all I wanted and there are some more things to refactor in the `EvidenceController`, but I think we can go ahead and review / merge this PR. It doesn't make things worse. I'll pick up further refactoring of the controller / form objects in a separate PR.

app/helpers/summary_helper.rb
```diff
@@ -12,10 +12,18 @@ def build_section(summary_text, object, fields)
...
     content_tag(:div, class: 'row') do
       content_tag(:div, class: 'small-12 medium-7 large-8 columns') do
         content_tag(:h4, "#{summary_name}")
+      end + build_link(link_title, link_url)
+    end
+  end
+
+  def build_link(link_title, link_url)
+    if link_title.present? && link_url.present?
+      content_tag(:div, class: 'small-12 medium-4 large-3 columns') do
```
colinbruce: I'd have set the classes to 'medium-5 large-4', that will balance with the foundation 12 columns 'medium-7 large-8' of the H4 row above

colinbruce:
Other than minor comment above, happy to merge.

app/helpers/summary_helper.rb
novotnyjakub: I had that in place, but it seems some of the elements margins extend the grid, so this was a lazy way of me not having to deal with it. But I'll give it a shot :) 

Good spot, btw!

colinbruce:
Go for merge! :checkered_flag:

PR #341, opened 2015-10-10
Renamed the app from the old to the new name

dotemacs:
Bug fix, as per Leigh's ticket:
https://www.pivotaltracker.com/story/show/105323540


colinbruce:
Looks good

PR #343, opened 2015-10-11
Removed unused DWP

dotemacs:
Removed all the tests and the classes in the app, as we're no longer
using this feature.

The main reason for doing this is that it just sat around and made the
output of the tests cluttered. Also there's a ticket in the backlog to
remove this...

The same as [PR339](https://github.com/ministryofjustice/fr-staffapp/pull/339), 
but instead of pointing it at **master** branch, I've pointed it at
**evidence-controller** as that is where the current development momentum is.


colinbruce:
As discussed, happy to merge, but there is more that can be removed! :+1:

PR #344, opened 2015-10-12
Added clearing of the reason

dotemacs:
This feature was covered by the feature test, but what wasn't covered is
the textarea wasn't being cleared after the users chose the 'Yes'
option.

Added the feature that clears the textarea and the test.

### DIFF
--- a/spec/javascripts/evidence_confirmation_spec.coffee
+++ b/spec/javascripts/evidence_confirmation_spec.coffee
@@ -70,8 +70,8 @@ describe "EvidenceConfirmationModule", ->
         @reason.append('EXPLANATION')
         @yes_button.trigger('click')
 
-      it 'removes the reason text', ->
+      it 'hides the reason section', ->
         expect(@reason_section.is(':visible')).toBe false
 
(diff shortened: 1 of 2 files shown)


spec/javascripts/evidence_confirmation_spec.coffee
```diff
@@ -0,0 +1,77 @@
+#= require radio_buttons_module
+#= require evidence_confirmation
+
+describe "EvidenceConfirmationModule", ->
+  element = null
+  beforeEach ->
+    element = $("""<div class="small-12 medium-8 large-5 columns">
```
novotnyjakub: I'm just thinking (and I don't know the answer), if there's a way of rendering this HTML from the actual codebase, rather than having it pasted here. Because if we change the markup for that form, this test may still pass.

colinbruce:
There is potential to use https://github.com/crismali/magic_lamp#magic-lamp to generate actual views rather than static html.  That is a refactoring task however, this can be merged and a separate ticket created for improvements.

PR #346, opened 2015-10-12
updated wording inline with design

colinbruce:
The design calls for the button to have a specific
wording, not just Next.

More work is needed to submit this page as a form to
set completed_at and _by values.  A separate branch
will be created for this.


dotemacs:
Great, merge at will

PR #348, opened 2015-10-13
Confirm evidence

colinbruce:
When the user clicks 'Complete processing' on the summary page, 
the model should record when it was clicked and by whom.

This adds the required fields, the routes and actions to handle it.

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -43,6 +43,10 @@ class EvidenceController < ApplicationController
     evidence_result
   end
 
+  def summary_save
+    record_confirmation
+  end
+
   def confirmation
     evidence_confirmation
   end
@@ -71,6 +75,14 @@ class EvidenceController < ApplicationController
     else
       redirect_to evidence_summary_path
     end
+  end
+
+  def record_confirmation
+    evidence.update(
+      completed_at: Time.zone.now,
+      completed_by: current_user
+    )
+    redirect_to evidence_confirmation_path
   end
 
   def income_params
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -107,6 +107,8 @@ ActiveRecord::Schema.define(version: 20151013143545) do
     t.string   "outcome"
     t.integer  "amount_to_pay"
     t.string   "incorrect_reason"
+    t.datetime "completed_at"
+    t.integer  "completed_by_id"
   end
 
   add_index "evidence_checks", ["application_id"], name: "index_evidence_checks_on_application_id", using: :btree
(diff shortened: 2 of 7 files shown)


GitCop:

 * Commit: 73539e4478e19682e6a326435fbf8394915a9553

---

spec/controllers/evidence_controller_spec.rb
```diff
@@ -167,6 +168,36 @@
...
+      expect(response).to have_http_status(200)
+    end
+
+    it 'renders the correct template' do
+      expect(response).to render_template :summary
+    end
+  end
+
+  describe 'POST #summary_save' do
+    let(:user)     { create :user, office: create(:office) }
```
colinbruce: @dotemacs In the evidence_controller_spec, I had to create the models
rather than stubbing, as there was no view model to mock and the 
current_user would need to be mocked and stubbed for one test.

Happy to discuss if you would rather these were stubbed, I wasn't sure on 
the time it would take/vs the time it would save!

GitCop:

 * Commit: c559279862bffdd5595c64d4e10964e7ae5ed902

---

dotemacs:
Sorry it took me so long to comment on this.

Fully approved!

db/schema.rb
```diff
@@ -98,17 +98,20 @@
...
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.datetime "expires_at",     null: false
+    t.datetime "expires_at",      null: false
     t.boolean  "correct"
     t.integer  "income"
     t.string   "outcome"
     t.integer  "amount_to_pay"
+    t.datetime "completed_at"
+    t.integer  "completed_by_id"
```
novotnyjakub: We briefly caught up with @dotemacs about this, because it seems unusual :) 

You're right, that it's a reference, so `_id` seems appropriate, but I haven't seen it used like this. I think the convention is usually just to use `_by`. `completed_by` is not a model anyway, so there needs to be a manual relationship setup in the `EvidenceCheck`. 

What do you think?

app/controllers/evidence_controller.rb
```diff
@@ -91,6 +95,14 @@ def income_form_save
     end
   end
 
+  def record_confirmation
+    evidence.update(
+      completed_at: Time.zone.now,
+      completed_by_id: current_user.id
```
novotnyjakub: I think we should setup `belongs_to` relationship in `EvidenceCheck` and then you can just pass the `current_user`, instead of directly manipulating id.

colinbruce:
@novotnyjakub Thanks for the feedback.  Amended as per your comments.  The DB field is still named completed_by_id by the belongs_to allows us to reference completed_by on the object.

novotnyjakub:
OK, makes sense, I think it's good to go :+1:

PR #349, opened 2015-10-13
Improve evidence controller and forms

novotnyjakub:
This is not just a refactoring PR, but also a preparation for the new Payment flow.

As discussed and agreed with @dotemacs, we'll for now put the `reason` back to the `EvidenceCheck` table. That way we should be able to re-use exactly the same form object for the new flow. We've briefly discussed and I'm sure there'll be more how to slice our models going forward.

It's also one further step to make our form objects as similar to each other as possible.

Last but not least the `EvidenceController` should now be cleaner and not have any responsibility for parameters processing.

### DIFF
--- a/spec/models/evidence/forms/income_spec.rb
+++ b/spec/models/evidence/forms/income_spec.rb
@@ -1,12 +1,13 @@
 require 'rails_helper'
 
 RSpec.describe Evidence::Forms::Income do
-  params_list = %i[id amount]
+  params_list = %i[income]
 
-  let(:amount) { '500' }
-  let(:income) { { id: 1, amount: amount } }
+  let(:evidence) { build_stubbed :evidence_check }
+  let(:income) { '500' }
+  let(:params) { { income: income } }
 
-  subject { described_class.new(income) }
+  subject(:form) { described_class.new(evidence) }
 
   describe '.permitted_attributes' do
     it 'returns a list of attributes' do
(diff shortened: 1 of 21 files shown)


dotemacs:
I guess this line can be taken out now:
https://github.com/ministryofjustice/fr-staffapp/blob/improve_evidence_controller_and_forms/app/controllers/evidence_controller.rb#L88 ?

spec/models/evidence/forms/income_spec.rb
```diff
@@ -15,45 +16,67 @@
...
+    it 'saves the income on the evidence in the correct format' do
+      subject && evidence.reload
+
+      expect(evidence.income).to eql(500)
     end
 
-    it 'saves the form data into appropriate models' do
-      expect(subject.save).to eq true
+    it 'saves the income on the income calculation outputs' do
+      subject && evidence.reload
```
dotemacs: Could the code on the lines 76 & 70 be put in a

```ruby
before(:each) { }
```

Not a show stopper, just saw it and figured the repetition could potentially be removed.

If this is the only thing in this PR, feel free to merge it and maybe revisit this on a next PR...

dotemacs:
Apart from the comments I made I see no problems with this PR.

And the comments I did make, none of them are to prevent this PR to be merged in.

Thanks, good work I especially like the specs which are not persisting the models but using stubs and mocks

PR #351, opened 2015-10-14
Updated evidence check letter

leighmoney:
Updated to include: “You should send this letter back with your
evidence” so staff can identify evidence for checking.


colinbruce:
Looks fine, merging.

PR #352, opened 2015-10-14
Evidence check creation fix

colinbruce:
Added evidence_check feature flag tests, then implemented 
a fix where the 10th income application would get flagged
even if the feature flag was turned off.


novotnyjakub:
:+1: I'll merge when the build gets green

PR #354, opened 2015-10-14
Implemented application emergency

dotemacs:
Some application's will need to be expedited through the system, so
they'll need to be declared as emergency applications.

This change allows for that.

Also, no emergency application will be selected for evidence check.

**WRONG THINGS:** currently in the ApplicationDetail model I'm assigning an
attribute called `emergency' which I manually remove in the Application
controller and in the ApplicationDetail model specs. The reason for that
is that currently Application model is one huge monolith and once we
split it up, these ugly hacks will be removed.

### DIFF
--- a/app/controllers/applications/build_controller.rb
+++ b/app/controllers/applications/build_controller.rb
@@ -79,6 +79,7 @@ class Applications::BuildController < ApplicationController
     if @form.valid?
       status = { status: get_status(step) }
       form_params.delete('application_id')
+      form_params.delete('emergency') if form_params.key?(:emergency)
       @application.update(form_params.merge(status))
       render_wizard @application
     else
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -197,9 +197,14 @@ RSpec.describe Application, type: :model do
       let!(:application_1) { create :application_part_remission }
       let!(:application_2) { create :application_full_remission }
       let!(:application_3) { create :application_no_remission }
+      let!(:emergency_application) { create :application_full_remission, emergency_reason: 'REASON' }
 
       it 'includes only part and full remission applications' do
         is_expected.to match_array([application_1, application_2])
+      end
+
+      it 'does not include emergency applications' do
+        is_expected.not_to include emergency_application
       end
     end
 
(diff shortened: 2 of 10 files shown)


app/controllers/applications/build_controller.rb
```diff
@@ -79,6 +79,7 @@ def handle_form_object(params, step)
     if @form.valid?
       status = { status: get_status(step) }
       form_params.delete('application_id')
+      form_params.delete('emergency') if form_params.key?(:emergency)
```
dotemacs: This is messed up, I know. But since we haven't broken up the **Aplication** model into smaller models just yet, this has to go in, for the form to be submitted.

spec/models/applikation/forms/application_detail_spec.rb
```diff
@@ -17,6 +19,7 @@
     let(:form) { described_class.new(application) }
 
     params_list.each do |attr_name|
+      next if attr_name.equal?(:emergency)
```
dotemacs: HACK, the same as the previous comment.

colinbruce:
Looks fine, noted on the caveat about the application model! :+1:

spec/models/application_spec.rb
```diff
@@ -194,10 +194,15 @@
       let!(:application_1) { create :application_part_remission }
       let!(:application_2) { create :application_full_remission }
       let!(:application_3) { create :application_no_remission }
+      let!(:emergency_application) { create :application_full_remission, emergency_reason: 'REASON' }
 
       it 'includes only part and full remission applications' do
         is_expected.to match_array([application_1, application_2])
       end
+
+      it 'does not include emergency applications' do
```
dotemacs: This @novotnyjakub

dotemacs:
@colinbruce & @novotnyjakub this branch wasn't passing the **Jiten test™** because it had a migration that that was merged into an older migration, so the migration from this branch never ran.

I've explicitly renamed the migration to be 'in the future' so that it gets picked up and runs.

Hopefully this will resolve the issue Jiten experienced: of never getting any evidence checks on applications.

dotemacs:
Closing this, to get the branch builder to rebuild it as it's not picking up the latest changes...

PR #355, opened 2015-10-15
Introduce payments

novotnyjakub:
This introduces the initial page of the new payments flow `/payments/:id`. The page can be only accessed by directly hitting the url as there's no link to it yet. Also the payment record has to be manually created in the database first for an application.

Apart from creating a new model and db table, this PR also significantly reuses the existing view models. It's a step to having a collection of those and being able to re-use them across all the 3 main journeys eventually. For now at least between evidence and payment flow.

I think it's ready to be merged to `master`, so we don't diverge much.


dotemacs:
@colinbruce I hope that this change will not 'mess with' your work too much.

@novotnyjakub reviewing now!

colinbruce:
LGTM. Merge at will.  

This might impact on the application summary page ticket that I'm looking at so I might have questions!

app/models/views/application_overview.rb
```diff
@@ -0,0 +1,53 @@
...
+        'none' => I18n.t('activerecord.attributes.application.summary.failed')
+      }[@application.application_outcome]
+    end
+
+    def savings
+      {
+        'true' => I18n.t('activerecord.attributes.application.summary.passed'),
+        'false' => I18n.t('activerecord.attributes.application.summary.failed')
+      }[@application.savings_investment_valid?.to_s]
+    end
```
dotemacs: This is a small suggestion. See what you think.

This method **savings** & the one above **income** potentially could work as:

```ruby
    def income
      format_locale @application.application_outcome
    end

    def savings
      format_locale @application.savings_investment_valid?.to_s
    end

    private

    def format_locale(value)
      prefix = 'activerecord.attributes.application.summary.'
      I18n.t([prefix, value].join)
    end
```

And then in the locale file add the appropriate keys:

```
activerecord.attributes.application.summary.(part|full|none|false|true)
```

What do we gain with this change? 
Looking at it now, the repetition would be moved from the codebase to the locale keys.

Also, now that I've thought about it a bit more, this might not work once we move away from the monolithic application model... The keys might have to be stored under a different hierarchy...

See what you think.

app/models/views/application_result.rb
```diff
@@ -0,0 +1,48 @@
...
+    private
+
+    def outcome
+      case evidence_or_application
+      when EvidenceCheck
+        evidence_or_application.outcome
+      when Application
+        evidence_or_application.application_outcome
+      end
+    end
```
dotemacs: @novotnyjakub you probably have this better in your head right now as you were knee deep in this code, so you might be able to answer this:

Could we potentially use duck typing here? Instead of doing this check to see if it's **EvidenceCheck** or **Application**, can't we just pass in an object instance, that answers to some method name that would produce the desired result e.g. @object.outcome ?

Or is this maybe too complex now to derive because we don't have the **Decision** model which could be queried?
Something like ```@application.decision.outcome```, which would be updated regardless if there was or wasn't an evidence check?

What are your thoughts?

dotemacs:
I really like the pull request and the changes it brings.
My comments above are just raised for discussion only!

app/models/views/application_overview.rb
novotnyjakub: I think it's sensible, this code looks a bit to verbose here. It's encapsulated in the view model, so not huge problem, but I'll try your suggestion..

app/models/views/application_result.rb
novotnyjakub: My thoughts were the same. The reason I did this is that I didn't want to worry about renaming or taking the decisions out. And at the moment we have an inconsistent naming in those 2 models. 

I'd suggest though to leave it for now (as it's a bit more complex to solve). Hopefully this will be the only view model we do it in for now and then once we create the `Decision` model and abstract things to it, things will be naturally cleaner.

Is that ok?

dotemacs: No objections from me. Just wanted to know the thought process behind it. Thanks :+1:

dotemacs:
Looks good to me, thanks

PR #356, opened 2015-10-15
Updated reset password email

colinbruce:
[delivers #103519722]

* updated wording in locale file
* changed html email to output url rather than link
* created plain text version of email

### DIFF
--- a/config/locales/devise.en.yml
+++ b/config/locales/devise.en.yml
@@ -23,7 +23,7 @@ en-GB:
         subject: "Please confirm your new password request"
         hello: "Dear %{name}"
         someone_requested_reset: "We received a request to change your password."
-        guide: "You can follow this link to "
+        guide: "To do this, copy and paste the link below into a firefox browser window: "
         link_text: "change your password."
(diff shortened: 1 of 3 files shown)


config/locales/devise.en.yml
```diff
@@ -23,7 +23,7 @@ en-GB:
         subject: "Please confirm your new password request"
         hello: "Dear %{name}"
         someone_requested_reset: "We received a request to change your password."
-        guide: "You can follow this link to "
+        guide: "To do this, copy and paste the link below into a firefox browser window: "
```
dotemacs: Maybe capitalise the word 'firefox' ? Ask @leighmoney to confirm.

The only reason I mention this is because the email will be all text so it would be good to make it obvious to use Firefox.

Or maybe even add **"(and not Internet Explorer)!"**

colinbruce: We are leaving as is for now.  @leighmoney has spoken! :)
Can I merge now?

dotemacs: Sure, go ahead :)

PR #357, opened 2015-10-15
Updated sign-up email

colinbruce:
[delivers #105118410]

* updated wording in locale file
* changed html email to output url rather than link
* created plain text version of email


dotemacs:
:green_heart:

colinbruce:
@dotemacs You'll be glad to know that @leighmoney has spoken again, and the FIrefox is now capitalised! :)

PR #358, opened 2015-10-15
Broken DB?

dotemacs:
Whenever I run the migrations this column gets added!

I can see that it was added in

20150508131347_add_fields_to_dwp_check.rb the commit is 9a2a8399.

Could it be that this was omitted during one of the merge
fiestas/fiascos?

Creating this so that we can discuss it!


dotemacs:
Removing this as it was only opened to start a discussion...

PR #359, opened 2015-10-15
Emergency application

dotemacs:
Some application's will need to be expedited through the system, so
they'll need to be declared as emergency applications.

This change allows for that.

Also, no emergency application will be selected for evidence check.

**WRONG THINGS:** currently in the ApplicationDetail model I'm assigning an
attribute called `emergency' which I manually remove in the Application
controller and in the ApplicationDetail model specs. The reason for that
is that currently Application model is one huge monolith and once we
split it up, these ugly hacks will be removed.


dotemacs:
This [build](https://travis-ci.org/ministryofjustice/fr-staffapp/builds/85541359) failed, with seed **31774**, tried to recreate it locally, no luck.

Will rebuild again to see how it fares. But there's obviously a race condition.

colinbruce:
Looks fine, when green.

colinbruce:
Still looking good.

PR #360, opened 2015-10-15
Add payments accuracy page

novotnyjakub:
**We need to sort out migrations order before merging this**

This adds the accuracy page to the payment flow. I needed to do a bit of refactoring on the evidence flow and the forms, but all components are reused between those.


dotemacs:
Looks good, merge it in

PR #361, opened 2015-10-16
Two digit year fix

colinbruce:
[finishes #104996152]
Addresses these [sentry alerts](https://sentry.service.dsd.io/mojds/prod-staff-app/group/247/)

When a user entered a two digit year into date of birth, a 500 error was
being generated.  This fix returns an error to the user and returns to
the page with a validation message.


dotemacs:
Looks fine to me

PR #362, opened 2015-10-16
Add values to letters

colinbruce:
[finishes #105467270]
[finishes #105467530]

The evidence check letters for part and no remission had 
placeholders for values.  Replaced the placeholders with 
the required data.


dotemacs:
Great, merge when :green_heart:

PR #363, opened 2015-10-16
Add payments to dashboard

novotnyjakub:
Again I tried to re-use as much as possible from the existing waiting for evidence table. Both are almost identical.

It also completes another story, which is changing the expiry format and order of the applications waiting for evidence / payments.


app/controllers/application_controller.rb
```diff
@@ -22,4 +22,10 @@ def after_sign_in_path_for(resource)
   def evidence_check_enabled?
     Settings.evidence_check.enabled == true
   end
+  helper_method :evidence_check_enabled?
+
+  def payment_enabled?
+    Settings.payment.enabled == true
```
dotemacs: Pedantic suggestion: 

```ruby
Settings.payment.enabled.equal?(true)
```

colinbruce:
Looks fine to me...

dotemacs:
Good to go!

PR #367, opened 2015-10-16
Payment confirmation

novotnyjakub:
The last bit of the payment journey.


dotemacs:
Looks fine to me.

@colinbruce shall we merge it now, during the weekend, so that it's fully ready to be tested on Monday?

colinbruce:
Why not?  Then we can work on creating the payments at the end of the application form and evidence check flows.

dotemacs:
Amazing!

PR #368, opened 2015-10-19
Removed superceded tests for application model

colinbruce:
As the application form is being refactored into form objects,
these tests have been duplicated in the new form object test
folder.

Having the duplication was causing un-necessary complexity in the
local file due to retaining the active record error messages.

This removal will make the refactoring simpler and more straight
forward.


dotemacs:
Good to go!

PR #369, opened 2015-10-19
Move locale to activemodel

colinbruce:
The application form referenced the local file in the ActiveRecord
namespace.  Once the form objects were created, the labels 
needed moving into the ActiveModel namespace.  This left 
duplication and some form labels/errors that had not transitioned 
properly.

This PR moves all entries into their proper namespaces and 
updates the savings_investment model to add the correct error
messages.


dotemacs:
When all the tests go green, merge at will

PR #370, opened 2015-10-20
Create part-payments

colinbruce:
When a user reaches the end of the application form and
evidence check processes, a payment record is created
when the result is a partial-remission.  

If an application form results in partial and is selected for
evidence check, the payment record is not created.

### DIFF
--- a/app/services/payment_builder.rb
+++ b/app/services/payment_builder.rb
@@ -12,8 +12,12 @@ class PaymentBuilder
 
   def part_payment?
     unless @application.payment?
-      @application.application_outcome == 'part' && !@application.evidence_check?
+      @application.application_outcome == 'part' && evidence_check_payment_validation?
     end
+  end
+
+  def evidence_check_payment_validation?
+    !@application.evidence_check? || @application.evidence_check.completed_at.present?
   end
 
   def expires_at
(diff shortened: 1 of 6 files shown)


colinbruce:
I had to turn off ClassLength rubocop on the Application model, this should be removed once the refactoring is in progress.

dotemacs:
Feel free to merge

app/services/payment_builder.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def decide!
+    @application.create_payment(expires_at: expires_at) if part_payment?
+  end
+
+  private
+
+  def part_payment?
+    unless @application.payment?
+      @application.application_outcome == 'part' && evidence_check_payment_validation?
```
dotemacs: PEDANTIC: there are two potential approaches here

a) one way of writing the above could be

```ruby
@application.application_outcome.equal?('part')
```

b) but a better way for me would be to take:

```ruby
@application.application_outcome == 'part' && evidence_check_payment_validation?
```
and move it to:

```ruby
def SOME_MEANINGFUL_NAME?
  @application.application_outcome.equal?('part') && evidence_check_payment_validation?
end
```

and then I'd re-write **part_payment?** as:

```ruby
def part_payment?
  SOME_MEANINGFUL_NAME? unless @application.payment?
end
```

PR #375, opened 2015-10-20
When no NI number provided, prevent loop

dotemacs:
There was an issue when a user didn't provide their NI number and the
application was started, the benefit check was attempted but it could
not be done because benefit check requires the NI number for the lookup.
This would throw the user into a loop. This change tries to prevent
that.

### DIFF
--- a/app/controllers/benefit_overrides_controller.rb
+++ b/app/controllers/benefit_overrides_controller.rb
@@ -1,0 +1,29 @@
+class BenefitOverridesController < ApplicationController
+  before_action :setup_application
+
+  def paper_evidence
+    @form = BenefitOverride.new(application_id: @application.id)
+  end
+
+  def paper_evidence_save
+    evidence = allowed_params['correct']
+    @form = BenefitOverride.find_or_create_by(application_id: @application.id)
+    @form.correct = evidence
+
+    if @form.save
+      redirect_to application_build_path(application_id: @application.id, id: :summary)
+    else
+      redirect_to paper_evidence_path(@application)
+    end
+  end
+
+  private
+
+  def setup_application
+    @application = Application.find(params[:application_id])
+  end
+
+  def allowed_params
+    params.require(:benefit_override).permit(*Forms::BenefitsEvidence.permitted_attributes)
+  end
+end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -109,7 +109,7 @@ en-GB:
     :no:
       heading: '✗ &nbsp; The applicant is not receiving benefits'
     details_missing:
-      heading: DWP check couldn't be performed, please make sure you've provided all personal details including National Insurance number
+      heading: "The applicant's details could not be checked with the Depatment for Work and Pensions"
     deleted:
       heading: The applicant is not receiving benefits
     deceased:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,25 +1,8 @@
-# Files in the config/locales directory are used for internationalization
-# and are automatically loaded by Rails. If you want to use locales other
-# than English, add the necessary files in this directory.
-#
-# To use the locales, use `I18n.t`:
-#
-#     I18n.t 'hello'
-#
-# In views, this is aliased to just `t`:
-#
-#     <%= t('hello') %>
-#
-# To use a different locale, set it with `I18n.locale`:
-#
-#     I18n.locale = :es
-#
-# This would use the information in config/locales/es.yml.
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
-
 en-GB:
+  benefit_override:
+    title: 'Do you have a letter confirming the applicant is receiving one of the benefits listed in question 9?'
+    incorrect: 'No'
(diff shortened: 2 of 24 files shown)


config/locales/en.yml
```diff
@@ -109,7 +92,7 @@ en-GB:
     :no:
       heading: '✗ &nbsp; The applicant is not receiving benefits'
     details_missing:
-      heading: DWP check couldn't be performed, please make sure you've provided all personal details including National Insurance number
+      heading: "The applicant's details could not be checked with the Depatment for Work and Pensions"
```
colinbruce: Department spelled wrong

spec/features/applications/benefit_results_are_rendered_spec.rb
```diff
@@ -19,6 +19,7 @@
     context 'after user selects yes to benefits' do
       context 'when NI has not been provided', focus: true do
         let(:ni_number) { nil }
+        let(:error_message) { "The applicant's details could not be checked with the Depatment for Work and Pensions" }
```
colinbruce: Wrong spelling of department here too.

app/controllers/benefit_overrides_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  private
+
+  def setup_application
+    @application = Application.find(params[:application_id])
+  end
+
+  def parse_benefit_override(params)
```
novotnyjakub: Why do we need more than, this? I'd try to avoid any params pre-processing in the controller. 

```ruby
params.require(:benefit_override).permit(*Forms:: BenefitsEvidence.permitted_attributes)
```

spec/features/applications/no_ni_number_and_undetermined_result_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+  scenario 'correct warning message' do
+    warning_string = "The applicant's details could not be checked with the Department for Work and Pensions"
+    expect(page).to have_content warning_string
+  end
+
+  scenario '"Next" button' do
+    expect(page).to have_button 'Next'
+  end
+
+  scenario 'link to paper evidence' do
```
novotnyjakub: This scenario is not needed, I think. We're testing in the `context` bellow, that user can click the link and the pre-conditions of the test are the same. 

I'm just thinking that it's not necessary to have a feature test for this, it'll slow down the suite. If you want to check that the link is there, maybe join to the `scenario 'correct warning message'`?

spec/features/applications/no_ni_number_and_undetermined_result_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    savings_and_investments
+    benefits_page
+  end
+
+  scenario 'correct warning message' do
+    warning_string = "The applicant's details could not be checked with the Department for Work and Pensions"
+    expect(page).to have_content warning_string
+  end
+
+  scenario '"Next" button' do
```
novotnyjakub: Not exactly same as the scenario bellow, but I'd maybe also join it with the previous one, just so we don't run 3 feature tests to verify content of the same page? I know that it's a good practice to have a single expectation per test, but feature tests are slow, so I'd consider to rename the above scenario to have a wider test coverage and put these in.

spec/features/applications/no_ni_number_and_undetermined_result_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      before do
+        choose 'benefit_override_correct_true'
+        click_button 'Next'
+      end
+
+      scenario 'takes them the summary page' do
+        expect(page).to have_content 'Check details'
+      end
+
+      scenario 'shows the full remission message' do
```
novotnyjakub: Same as the comments above, I'd join these scenarios to 1, to save time.

spec/features/applications/when_benefits_checker_returns_undetermined_result_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  include Warden::Test::Helpers
+  Warden.test_mode!
+
+  before do
+    dwp_api_response 'Undetermined'
+
+    login_as user
+    visit applications_new_path
+
+    drive_to_the_benefits_page
```
novotnyjakub: Is there a benefit of navigating through all the pages, instead of using a factory and then loading the benefits page?

spec/features/applications/no_ni_number_and_undetermined_result_spec.rb
dotemacs: Fair point. Removing it now.

spec/features/applications/no_ni_number_and_undetermined_result_spec.rb
dotemacs: Looking from the angle that you're looking at it, good observation. But I had it in, as I was developing the feature by feature, going from the outside.

Fine, I'll look at taking this out.

spec/features/applications/when_benefits_checker_returns_undetermined_result_spec.rb
dotemacs: Yea, it's a end to end test.

PR #381, opened 2015-10-21
Refactor Summary form object

colinbruce:
Long was the day and heavy was my heart as I finally pushed code that worked.

This PR (finally) moves the Summary and Confirmation pages from the (soon to be refactored out of existence) build_controller into the shiny new process controller.

It required a new result view model (although it will probably be a target for refactoring itself in the future) because it needs a different set of fields than the existing evidence_check and payment result models.

Where I found them, I moved the tests covering the build controller into the new namespace, but there may be some future refactoring work to ensure that the features and controller tests are all mapping to the right areas.


app/controllers/applications/process_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Applications
   class ProcessController < ApplicationController
     before_action :authenticate_user!
+    before_action :evidence_check_show_redirect, only: :confirmation
```
novotnyjakub: Considering we now don't have Wicked in place and have control over each individual action, wouldn't it be more transparent to make the redirect in the `confirmation` action? I personally think that filters, which are part of the action's flow, loading data, etc. are always quite hidden and it's difficult to understand what the action itself does.

I'm not sure, what do you think?

app/models/views/application_overview.rb
```diff
@@ -2,8 +2,8 @@ module Views
   class ApplicationOverview
```
novotnyjakub: This class is getting quite big now, we should probably consider splitting it as well as it seems to have more responsibilities already. Not now, of course, that's just a thought.

app/models/views/applikation/result.rb
```diff
@@ -0,0 +1,28 @@
+module Views
+  module Applikation
```
novotnyjakub: I think we don't have to use this incorrect name any more, we can use `Application` as it's nested in the `Views` namespace, so it doesn't clash with any mode.

app/models/views/applikation/result.rb
```diff
@@ -0,0 +1,28 @@
...
+    class Result < Views::ApplicationResult
```
novotnyjakub: I'm wondering if it's not confusing which class has responsibility for which part, we have `Views::ApplicationResult` and then `Views::Application::Result`. I think I understand the difference, but it's not obvious from those names and I think it'll confuse us later. 

What about renaming the `Views::AppicationResult` to `Views::ApplicationDecision` in preparation to the decision setup we'll be adding later?

app/views/applications/process/summary.html.slim
```diff
@@ -0,0 +1,19 @@
+.row
```
novotnyjakub: This template looks so good now, great! You can now easily understand what's being rendered here.

app/controllers/applications/process_controller.rb
```diff
@@ -17,15 +18,35 @@ def personal_information_save
       end
     end
 
+    def summary
+      @result = Views::Applikation::Result.new(application)
+      @overview = Views::ApplicationOverview.new(application)
+    end
+
+    def summary_save
```
novotnyjakub: I would remove this action for now, until we need it. I think the `Complete processing` link should direct straight to the confirmation page for now, because it's just an extra redirect at this stage.

app/models/views/applikation/result.rb
CeeBeeUK: I'm pretty sure we can't Application seems to be a reserved word! :(

novotnyjakub: It can't be, we're using `Application` model all over. 

I've just tried to create this locally and it works: 
```ruby
module Views
  module Application
    class Something
    end
  end
end
```

app/models/views/applikation/result.rb
CeeBeeUK: Agreed, but maybe raise it as an issue and do it as a separate PR?  Looking at it, this will add an extra 9 files to this PR because of the model inheritance and subsequent tests

novotnyjakub: Of course, that was more of a discussion starter.

app/models/views/applikation/result.rb
CeeBeeUK: Weird, because everytime I have tried to rename this object to 
```ruby
module Views
  module Application
    class Result < Views::ApplicationResult
```
It bombed out... badly. Never mind tonight, I'll start again in the morning.

GitCop:

 * Commit: 087b968b85bec450296f299b7c484c6ca41861c5

---

PR #388, opened 2015-10-23
Personal information slice

novotnyjakub:
This is a first step to slimming down the monolithic `Application` model and also making `BuildController` much simpler. 

1. Personal information has been taken out of `Application` to `1-1` model called `Applicant`. Because of the various cross dependencies, I decided to build the `Applicant` model always together with `Application`. That's mostly to keep the current functionality working, but we'll refactor it piece by piece later, so that's created on demand.

2. `PersonalInformation` form object now accepts `Applicant` model, rather than `Application`. We should discuss what the forms naming convention should be.

3. There's a new `Applications::ProcessController`, which is like the _newer version of_ `Applications::BuilderController`. The idea is to port action by action and then decommission the `BuildController` altogether.


app/models/application.rb
```diff
@@ -33,27 +34,18 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
...
-  before_validation :format_ni_number
   before_validation :nullify_blank_emergency_reason
 
-  # Step 1 - Personal detail validation
-  with_options if: proc { active_or_status_is? 'personal_information' } do
-    validates :last_name, presence: true
-    validates :married, inclusion: { in: [true, false] }
-    validates :last_name, length: { minimum: 2 }, allow_blank: true
-    validates :date_of_birth, date: true
-    validate :dob_age_valid?
```
colinbruce: If these are deleted here, where are they being validated now?  https://github.com/ministryofjustice/fr-staffapp/pull/388/files#diff-1e3f6eb69dfed2e3702dad512eb74667R6 only seems to validate the NI Number below.

novotnyjakub: It's validated in the `Applikation::Forms::PersonalInformation` form object.

dotemacs:
Looks fine as the part of the ongoing Application model refactoring marathon.

PR #389, opened 2015-10-24
Extracted scopes from Application into Query objects

dotemacs:
This is just a discussion starter, to see what you @colinbruce & @novotnyjakub think.

This is just an example of how we can do things: instead of having it
all stuffed into the Application model, tease it out.

The new nested structure directory 'query', that is open for debate,
maybe it should be called 'queries'?

Also something like this can be extracted out into this new query object and tested tighter:
https://github.com/ministryofjustice/fr-staffapp/blob/extract-queries/app/services/evidence_check_selector.rb#L33

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -42,7 +42,7 @@ class HomeController < ApplicationController
   end
 
   def waiting_for_evidence
-    current_user.office.applications.waiting_for_evidence
+    Query::WaitingForEvidence.new(current_user).find
   end
 
   def waiting_for_payment
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -46,7 +46,7 @@ class HomeController < ApplicationController
   end
 
   def waiting_for_payment
-    current_user.office.applications.waiting_for_payment
+    Query::WaitingForPayment.new(current_user).find
   end
 
   def load_graph_data
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -11,14 +11,6 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
 
   validates :reference, presence: true, uniqueness: true
 
-  scope :waiting_for_payment, lambda {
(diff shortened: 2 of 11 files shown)


dotemacs:
The idea is to also extract, in the same way, the following methods:

https://github.com/ministryofjustice/fr-staffapp/blob/master/app/models/application.rb#L19
https://github.com/ministryofjustice/fr-staffapp/blob/master/app/models/application.rb#L27

dotemacs:
The failed test is RuboCop style violation: Line is too long.

dotemacs:
The rest is all :green_heart:

dotemacs:
Fixed the RuboCop issue.

The idea is to change this: https://github.com/ministryofjustice/fr-staffapp/blob/extract-queries/app/services/evidence_check_selector.rb#L33-L36

To something like:

```ruby
Query::EvidenceCheckable.new.find_all.refundable(@application.id, refund).count
```

dotemacs:
Just as I mentioned previously, this is just a discussion starter. All the names are up for debate!

novotnyjakub:
I would personally not merge this now, while we're slicing up the `Application` model. Once we're finished, we may have better idea about what else needs to be taken out.

dotemacs:
I think by taking this out, we can concentrate on the application model.
And potentially all the other scopes.

We would also reduce the Code Climate warnings about code duplication.

colinbruce:
I'm not sure that the evidencecheckable needs to be removed.  
The other two scopes are near duplicates
 but evidencecheckable is a single use scope.

dotemacs:
It doesn't 'need' to be removed because there are no alerts for it.

But it should be moved because it violates SRP and it also slims down the Application model.

dotemacs:
Removed all three scopes and switched them for query objects.

Please re-review again.

Thanks

colinbruce:
This definitely reduces the footprint of the Application model, that can only be good! :)

Happy to merge and see what it does to Code Climate :+1:

PR #391, opened 2015-10-26
Remove duplicated scope

colinbruce:
Each model had a check_by_day scope, this already existed in
the CommonScope object.  I removed the Object specific scope
and ensured CommonScope was included.


dotemacs:
Do it!

PR #392, opened 2015-10-28
Released fix for leap year bug

colinbruce:
* refactored into the applicant model
* test for leap year correct response
* test for delegation to applicant model

[ticket](https://www.pivotaltracker.com/story/show/106838812)
[finishes #106838812]

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -39,6 +39,7 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
   APPLICANT_SETTERS = %i[title= first_name= last_name= date_of_birth= ni_number= married=]
   delegate(*APPLICANT_GETTERS, to: :applicant)
   delegate(*APPLICANT_SETTERS, to: :applicant)
+  delegate(:age, to: :applicant, prefix: true)
 
   MAX_AGE = 120
   MIN_AGE = 16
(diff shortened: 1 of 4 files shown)


spec/models/application_spec.rb
```diff
@@ -245,6 +247,20 @@
     end
   end
 
+  describe '#age' do
```
novotnyjakub: We should test this method in `Applicant` model.

spec/models/application_spec.rb
```diff
@@ -245,6 +247,20 @@
...
+    subject { application.applicant_age }
+
+    context 'when applicant is born on Feb 29th in a leap year' do
```
novotnyjakub: Is one single scenario enough for all cases? That's just a question, I'm not sure.

app/models/application.rb
```diff
@@ -151,18 +152,13 @@ def full_name
   end
 
   def applicant_over_61?
-    applicant_age >= 61
+    applicant.age >= 61
```
novotnyjakub: I'm thinking whether we do use the `age` somewhere outside this `applicant_over_61?` method. If not, than we don't actually need the delegation. I thought originally that we use the age directly.

colinbruce: We are using it a couple of places; a partial view and a service object.

spec/models/application_spec.rb
colinbruce: Noted.  I created the test before the refactoring work.

app/models/application.rb
novotnyjakub: :+1:

PR #393, opened 2015-10-28
Enabled caching of gems for Travis CI

dotemacs:
This parameter enables caching of gem for Travis CI, so it speeds up the
testing.

Travis CI docs here:
http://docs.travis-ci.com/user/caching/


colinbruce:
:+1: Definitely, can't hurt!

PR #395, opened 2015-10-29
Update personal_information form

colinbruce:
This will reduce the code duplication warning in coedclimate
and increase the file from an B to an A, an improve the GPA


dotemacs:
Looks good

PR #397, opened 2015-10-29
Fix for part payments from evidence-checks

colinbruce:
Implement test for creation of object at end of evidence-check flow
Update the payment builder to check for application and
evidence check being passed in.

The Payment builder was only looking at the application model to check
for part payment results, I re-used the code from the application_result
view model to test for the presence of both.

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -88,7 +88,7 @@ class EvidenceController < ApplicationController
       completed_by: current_user
     )
     if payment_enabled?
-      PaymentBuilder.new(@evidence.application, Settings.payment.expires_in_days).decide!
+      PaymentBuilder.new(@evidence, Settings.payment.expires_in_days).decide!
     end
     redirect_to evidence_confirmation_path
   end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -94,6 +94,8 @@ class Application < ActiveRecord::Base # rubocop:disable ClassLength
   end
   # End step 5 validation
 
+  alias_attribute :outcome, :application_outcome
+
   def children=(val)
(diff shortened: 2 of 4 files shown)


novotnyjakub:
I was looking at this and thinking that the `PartPayment` builder should maybe accept either `Application` or `EvidenceCheck` in the constructor, based on which flow are you at. And then from the object's outcome, derive whether to create or not the `Payment` model. 

I'd also suggest for simplification to create a method alias from `Application::application_outcome` to `Application::outcome`, which would make this and the `ApplicationResult` objects much easier, because they wouldn't have to choose a method based on the object class. It'll help us with consistency and then later when we introduce decision, I think.

What do you think?

colinbruce:
I like the sound of it, but the payment object is always created on the application, not on the evidence check.  Because of that, I thought that this procedure would be ok, either way we have to derive the outcome 

The method alias sounds like a good idea too and might make the above easier.

novotnyjakub:
Were: https://github.com/ministryofjustice/fr-staffapp/blob/master/app/controllers/evidence_controller.rb#L91 we can pass `@evidence` instead of `@evidence.application`, can't we? 
The good thing is that we're getting out of the grip of the `Wicked` flow, so we have a better control over the whole flow.

novotnyjakub:
Cool, I didn't realise that we still need some kind of branch condition, because the `Payment` belongs to the `Application`. 

Let's merge this now, but I have an idea of how to maybe deal with this. I was thinking about adding payment to `EvidenceCheck` as well: 

```ruby
has_one :payment, through: :application
```

Because then we could simplify this class to remove `load_application` and use: 

```ruby
@initiator.create_payment(expires_at: expires_at) if part_payment_needed?
```

PR #398, opened 2015-10-29
Fix for RuboCop violation

dotemacs:
This is the failure:

https://travis-ci.org/ministryofjustice/fr-staffapp/builds/88100149

and this change fixes it.


dotemacs:
Merged the change when the PR went green. But it was only via email that I found out that the build failed RuboCop's check.

colinbruce:
:+1:  go when green

PR #402, opened 2015-10-30
Remove standalone calculator

colinbruce:
As part of the refactoring we need to remove the standalone r2 calculator 
and DWP checks.  This is the first part of this [ticket](https://www.pivotaltracker.com/story/show/104535314).


novotnyjakub:
:+1: best PR ever!

PR #403, opened 2015-10-30
Slice application details

novotnyjakub:
Second part of the big refactor of the `Application` model into a multiple separate ones. Again, there's a few hacks in the code, which have to be there for the moment. 

A big one is in `ProcessController` in the `application_details_save` action, which has to touch the `Application` for the `after_save` callback to run. This has to be there until we extract the benefits and income calculations from the model.


dotemacs:
Merge away!

PR #406, opened 2015-11-02
Remove dwp_check **blocked**

colinbruce:
As part of [this ticket](https://www.pivotaltracker.com/story/show/104535314) we are removing
the no longer used features.

This PR removes the stand-alone dwp_check feature.

**It is currently blocked because a data-retention policy is needed.**


spec/controllers/home_controller_spec.rb
```diff
@@ -28,9 +28,8 @@
 
       context 'as an admin' do
         before(:each) do
-          DwpCheck.delete_all
           Office.delete_all
-          FactoryGirl.create_list :dwp_check, 2, created_by: manager, office: manager.office
+          FactoryGirl.create_list :benefit_check, 2, user_id: manager.id
```
novotnyjakub: We're usually skipping `FactoryGirl` namespace here as the helpers are included in the specs. Like bellow.

novotnyjakub:
:+1: Looks good, I'm looking forward to merging this :)

PR #409, opened 2015-11-02
Refactor the application model

colinbruce:
When the personal_information model was split out, the age
validation code was [moved over](https://github.com/ministryofjustice/fr-staffapp/blob/master/app/models/applikation/forms/personal_information.rb#L56), along with the tests.

The original code was left behind and is no longer used.

This removal should improve the code-climate score for the
(decreasingly) monolithic application model.


dotemacs:
Sorry, yes, saw it, good to go!

PR #411, opened 2015-11-03
Refactored waiting_for queries

colinbruce:
Implemented a base class with the shared code.
Each query calls the super method using the 
symbol for the required attribute and extracting the 
database field name in the method.


dotemacs:
Looks good.

novotnyjakub:
:+1:

PR #415, opened 2015-11-03
fixed weird layout

Demwunz:
@kelliematheson


kelliematheson:
Tested on branchbuilder and accept

PR #416, opened 2015-11-03
Bug fix: prevent empty paper evidence submission

dotemacs:
When the Benefits Override flow's evidence page would be submitted
without the either options chosen, the form could be submitted and an
error would be thrown.

This commit prevent's the form to be submitted for saving without one of
the options being picked.

### DIFF
--- a/app/models/forms/benefits_evidence.rb
+++ b/app/models/forms/benefits_evidence.rb
@@ -7,8 +7,13 @@ module Forms
     define_attributes
 
     validates :correct, inclusion: { in: [true, false] }
+    validate :isnt_blank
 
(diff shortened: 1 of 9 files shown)


app/models/forms/benefits_evidence.rb
```diff
@@ -7,9 +7,14 @@ def self.permitted_attributes
     define_attributes
 
     validates :correct, inclusion: { in: [true, false] }
+    validate :isnt_blank
```
colinbruce: Looks fine, but wouldn't 
  validates :correct, presence: true 
have done the same thing?

PR #418, opened 2015-11-03
Allow the users to override benefits check

dotemacs:
Allow the staff to override benefit check when the result is 'No', if
the member of public presents paper based evidence.


novotnyjakub:
PR ok, but the added test is failing.

novotnyjakub:
:+1:

PR #420, opened 2015-11-03
updated guidance

Demwunz:
@kelliematheson


kelliematheson:
Tested on Branchbuilder and accept

PR #421, opened 2015-11-04
Reverted the validation changes

colinbruce:
The new time constants were being set on first creation and locking the date
to the server launch date.

Added a Timecop test to ensure that any future attempts to refactor this
don't break it again!


dotemacs:
Approved

PR #422, opened 2015-11-04
Re-refactor application_detail

colinbruce:
Following this morning's quick reversion of the previous refactoring 
code, this is another attempt at refactoring the application detail to 
remove the duplication.

Interested to discuss the pros and cons of the following
* 3.months.ago vs Time.zone.today - 3.months
* TIME_LIMIT_FOR_PROBATE.years.ago vs
 Time.zone.today - TIME_LIMIT_FOR_PROBATE.years

### DIFF
--- a/app/models/applikation/forms/application_detail.rb
+++ b/app/models/applikation/forms/application_detail.rb
@@ -29,26 +29,38 @@ module Applikation
       validate :emergency_reason_size
 
       validates :date_received, date: {
-        after: proc { Time.zone.today - 3.months },
-        before: proc { Time.zone.today + 1.day }
+        after: :min_date,
+        before: :tomorrow
       }
 
       with_options if: :probate? do
         validates :deceased_name, presence: true
         validates :date_of_death, date: {
-          after: proc { Time.zone.today - TIME_LIMIT_FOR_PROBATE.years },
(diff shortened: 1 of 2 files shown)


novotnyjakub:
I'm thinking, considering the bug yesterday, which our tests didn't pick up. We should probably have some tests around this, probably using timecop so that we can freeze time for a particular date and then test the variants. The problem is that our tests use `Time.today`, so they always pass.

colinbruce:
@novotnyjakub Like this? 
https://github.com/ministryofjustice/fr-staffapp/blob/master/spec/models/applikation/forms/application_detail_spec.rb#L330  I added one this morning on the reversion to make sure that I was reverting it properly!

novotnyjakub:
I meant more in the `describe 'validations' do` section, for each individual test. To freeze date for ie. `10/10/2015` and then test that `11/07/2015` is valid, `10/07/2015` invalid, etc. Because using `Time.zone.now - 3.months` is essentially the code, which means the is testing the implementation, rather than the conditions. 

Hope that makes sense.

colinbruce:
Okay, I see what you mean.

colinbruce:
@novotnyjakub can you re-review? Thanks.

PR #425, opened 2015-11-05
[DO NOT MERGE] Pundit example for processed applications

novotnyjakub:
This PR is not to be merged, but it's a conversation starter about authorisation. 

I wanted to demonstrate how we could use `pundit` to test-drive policies, which I'm not sure if it's easily possible in `cancancan`. Also the concept of policy_scope which filters records not complying with the policy is very usefull as well.

It's particularly cool to generate a full documentation of who can do what:

```ruby
rspec -f d spec/policies/application_policy_spec.rb
```

@dotemacs @colinbruce what do you think?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'devise'
 gem 'devise_invitable'
 # roles
 gem 'cancancan', '~> 1.10'
+gem 'pundit', '~> 1.0'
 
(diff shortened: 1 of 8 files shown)


novotnyjakub:
This example is no longer needed and the work is being carried on in a different branch

PR #427, opened 2015-11-06
Part of change from payments to part-payments

leighmoney:


colinbruce:
@leighmoney tests are fixed, you'll need to re-sync before you do any further work on this branch.

PR #428, opened 2015-11-06
Add completion fields to appliction

colinbruce:
This will allow us to record the user that completes the
application form.

It also aligns the models with similar properties, allowing
us to share code to set the attributes.

Setting the values will be implemented by the up-coming 
ResolverService

This will be needed for the manager dashboard changes at [ticket](https://www.pivotaltracker.com/story/show/107484086)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151030164736) do
+ActiveRecord::Schema.define(version: 20151106120125) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -54,6 +54,8 @@ ActiveRecord::Schema.define(version: 20151030164736) do
     t.integer  "amount_to_pay"
     t.boolean  "high_threshold_exceeded"
     t.string   "reference"
+    t.datetime "completed_at"
+    t.integer  "completed_by_id"
   end
 
   add_index "applications", ["office_id"], name: "index_applications_on_office_id", using: :btree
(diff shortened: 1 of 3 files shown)


colinbruce:
Questions:

* Should we code for completed_by being empty or should we migrate,
    * created_by to completed_by?
    * created_on to completed_at?
* Update query/processed_applications to look at where completed_at NOT nil?

db/schema.rb
```diff
@@ -31,12 +31,6 @@
   add_index "applicants", ["application_id"], name: "index_applicants_on_application_id", using: :btree
 
   create_table "applications", force: :cascade do |t|
-    t.string   "title"
-    t.string   "first_name"
-    t.string   "last_name"
-    t.date     "date_of_birth"
-    t.string   "ni_number"
-    t.boolean  "married"
     t.string   "status"
```
colinbruce: I've just noticed that these field have been deleted from my schema, despite me not touching them... I'm going to update the file and force push this branch.

novotnyjakub:
So there are no `created_on` and `created_by` fields in the db ;-) But it's a very good point and I think we should try to migrate if we can. 

But I'd make sure that we only migrate the ones which have `application_outcome` set and which have `status` set to `active` (or whichever represents the application has been completed). 

For the applications with that condition matching, I think we can migrate 
* `created_at` -> `completed_at`
* `user_id` -> `completed_by`

I'd even consider to later remove the `user_id` column as is, because it won't have much value.

colinbruce:
Looking at this now and we have an issue... the status column was used (and set) by wicked for the wizard.  Consequently it now doesn't reach, or pass the summary stage.  I'm trying to work out an algorithm now. :(

novotnyjakub:
So my main concern was not to migrate the 'zombie' applications, but maybe we can assume that if it has income and outcome setup, it's `completed`?

PR #429, opened 2015-11-06
Cleanup factories and improve stubbing

novotnyjakub:
This may not seem as an improvent, but it allows us to build stubbed
Application with Applicant / Detail. Before it was possible but without
being able to inject any of those when stubbing.

### DIFF
--- a/spec/factories/applicants.rb
+++ b/spec/factories/applicants.rb
@@ -1,9 +1,5 @@
 FactoryGirl.define do
   factory :applicant do
-    transient do
-      application nil
-    end
-
     factory :applicant_with_all_details do
       title { Faker::Name.prefix }
       first_name { Faker::Name.first_name }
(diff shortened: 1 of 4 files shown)


colinbruce:
I'm going to nod and smile , that  is incredibly confusing, but everything is green!
Merge at will! :)
:+1:

novotnyjakub:
I know it is.. I'm not happy about it either. 

In fact it got me thinking (more than once) if we made the right choice by splitting the `applications` database table. Because having lots of `1-1` tables, which require for all the records to exist, make no sense. Taking stuff from the `Application` model is one thing, but having a wide database table is not a performance issue. So I think we should discuss it and potentially re-consider.

colinbruce:
Okay.  I think the personal information was worth splitting out as, potentially, a single person could apply multiple times.  Less sure about the application details.  Happy to chat next week.

PR #431, opened 2015-11-06
Updated processed application display

colinbruce:
This will allows us to record the user that completes the application form by calling the new ResolverService

This PR delivers the manager dashboard changes at [ticket](https://www.pivotaltracker.com/story/show/107484086).

It aligns the Application model with those of similar properties, allowing us, in the future, to call a single point to set the completion attributes for Applications, Evidence_checks and Payments.

**Potential** 
After implementing a state-machine solution, the ResolverService could also act as a method for setting application decision and transitioning through the states.


config/routes.rb
```diff
@@ -16,6 +16,7 @@
     put 'application_details',
       to: 'applications/process#application_details_save', as: :application_details_save
     get 'summary', to: 'applications/process#summary', as: :summary
+    get 'summary_save', to: 'applications/process#summary_save', as: :summary_save
```
novotnyjakub: This should be a 'put' as it's updating the application.

app/views/applications/process/summary.html.slim
```diff
@@ -17,4 +17,4 @@ header
     .row.collapse.medium-10.large-8
       =render(partial: 'shared/remission_type', locals: { source: @result })
 
-=link_to 'Complete processing', application_confirmation_path(@overview.application.id), class: 'button primary'
+=link_to 'Complete processing', application_summary_save_path(@overview.application.id), class: 'button primary'
```
novotnyjakub: Related to the comment in `routes.rb`, this should be a form, even though an "empty one". A simple 

```
= form_for @application, url: :summary_save_path, method: :post, html: { autocomplete: 'off' } do |f|
  = f.submit 'Complete processing', :class => 'button primary'
``` 

We don't need form object for it or anything.

db/migrate/20151106120125_add_completed_to_application.rb
```diff
@@ -0,0 +1,22 @@
...
+    add_reference :applications, :completed_by, references: :users
+    migrate_sql = <<-SQL.gsub(/^\s+\|/, '')
+      |UPDATE applications
+      |SET
+      |  completed_at = "applications"."updated_at",
+      |  completed_by_id = "applications"."user_id"
+      |FROM "applications"
+      |WHERE application_outcome IS NOT NULL AND
+      |  status IS NOT NULL AND
+      |  application_type IS NOT NULL;
```
novotnyjakub: Yeah, I think this is sufficient. Maybe one or two will slip in which shouldn't be there, but that's fine :)

spec/controllers/applications/process_controller_spec.rb
```diff
@@ -144,6 +144,36 @@
     end
   end
 
+  describe 'GET #summary_save' do
+    let(:current_time) { Time.zone.now }
+    let(:user) { create :user }
+    let(:application) { create :application_full_remission }
+
+    before do
+      allow(Application).to receive(:find).with(application.id).and_return(application)
```
novotnyjakub: You don't have to do this if you're creating the application above (line 150). This is needed only if we use `build_stubbed` for the main model, because then it's not stored in the db. If you could though change it to `build_stubbed`, that's better I guess.

spec/controllers/applications/process_controller_spec.rb
```diff
@@ -144,6 +144,36 @@
...
+    it 'returns the correct status code' do
+      expect(response).to have_http_status(302)
+    end
+
+    it 'redirects to the confirmation page' do
+      expect(response).to redirect_to(application_confirmation_path(application.id))
+    end
+
+    it 'updates the payment completed_at and completed_by' do
+      application.reload
```
novotnyjakub: Same here, if you changed the `application` to be build as stubbed, then you don't have to reload it here.

novotnyjakub:
I'm seeing a bright and great future for this service :-) Good stuff! :+1:

colinbruce:
Cheers, looking at it now...

novotnyjakub:
:+1: Looks good!

app/services/resolver_service.rb
```diff
@@ -0,0 +1,21 @@
+class ResolverService
+  def initialize(object, user)
+    @calling_object = object
+    @user = user
+  end
+
+  def process
+    mark_complete
+    # TODO: implement evidence_check create for applications
+    # TODO: implement payment creation for applications and evidence_checks
```
dotemacs: End of the day, my brain is fried. Is this a reminder (the TODOs) for a future reference or for something that was meant to be added to this PR but was omitted ?

colinbruce: It was a placeholder for future use, in case I forgot where it would be! :)

spec/services/resolver_service_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:user) { build_stubbed(:user) }
+
+  subject(:resolver) { described_class.new(object, user) }
+
+  describe '#process' do
+    Timecop.freeze(Time.zone.now) do
+      before do
+        resolver.process
+      end
+      context 'when created with an application' do
```
dotemacs: this could be just a single line:

```ruby
before { resolver.process }
```

plus a white space could be added :neckbeard:

dotemacs:
@novotnyjakub can you please give this PR one more check.

I've just merged master into it. So if it's looking well, please update the conversation.

dotemacs:
@novotnyjakub :clock530:

PR #432, opened 2015-11-06
Fix processed application detail page for benefit based application

novotnyjakub:
There was a bug which tried to format income even though it was `nil`. That's now fixed.


novotnyjakub:
I'm merging this as it's a relatively urgent bug fix.

PR #433, opened 2015-11-07
Remove auto checks from Application

novotnyjakub:
A big refactoring piece which aims to remove the after save callbacks from `Application` model. 

* benefit check - :white_check_mark: 
* income - :white_check_mark: 

I've discovered some problems which I haven't fixed (no time or discussion needed):
* there's no validation for the `income` field, when saved without there is no error but user can't progress
* once `EvidenceCheck` or `Payment` are created, if the application is subsequently changed to one which shouldn't be checked, these records are still there
* not sure if there's a feature test coverage for creating `Payment` for `Application` when it's not evidence checked


app/controllers/applications/build_controller.rb
```diff
@@ -48,8 +53,8 @@ def show # rubocop:disable CyclomaticComplexity
   end
 
   def update
-    evidence_check_selection
-    create_payment_if_needed
+    # evidence_check_selection
+    # create_payment_if_needed
```
dotemacs: I'm guessing these two lines can be removed now?

app/controllers/applications/process_controller.rb
```diff
@@ -34,6 +35,67 @@ def application_details_save
...
+      if application.benefits
+        @application = application
+        render :benefits_result
+      else
+        redirect_to(application_build_path(application_id: application.id, id: :income))
+      end
+    end
+
+    def income
+      if !application.benefits?
```
dotemacs: Could we maybe add `.income?` to Application model, so that this reads 'better'?

It would essentially be the same as above:

```ruby
def income?
  !benefits?
end
```

app/controllers/applications/process_controller.rb
```diff
@@ -34,6 +35,67 @@ def application_details_save
...
+        calculate_income
+        evidence_check_and_payment
+        redirect_to(application_build_path(application_id: application.id, id: :income_result))
+      else
+        render :income
+      end
+    end
+
+    def income_result
+      if application.application_type == 'income'
```
dotemacs: And then maybe we could even re-use `.income?` here

app/services/benefit_check_runner.rb
```diff
@@ -0,0 +1,82 @@
...
+      detail.date_fee_paid
+    elsif detail.date_received.present?
+      detail.date_received
+    end
+  end
+
+  def build_hash
+    parts = [applicant.last_name, applicant.date_of_birth, applicant.ni_number, benefit_check_date]
+    Base64.encode64(parts.to_s)
+  end
```
dotemacs: We could potentially avoid having to use the temporary variable:

```ruby
  def build_hash
    Base64.encode64 [applicant.last_name,
                     applicant.date_of_birth,
                     applicant.ni_number,
                     benefit_check_date].to_s
  end
```

app/controllers/applications/process_controller.rb
dotemacs: I just had a look at db schema and Application already has income attribute, so I'm not sure if the above would work.

Maybe adding something like:

```ruby
def income_based?
  !benefits?
end
```

dotemacs: Just discussed this with @colinbruce, please disregard the comment about this.

novotnyjakub: Yeah, I agree with you that we need a cleaner way, but I think it'll be easier after we've done the state machine piece.

app/controllers/applications/build_controller.rb
novotnyjakub: Most of the controller code is not used any more, so I didn't bother cleaning it up. I'm hoping to remove it all together later today.

app/controllers/applications/process_controller.rb
novotnyjakub: Same as above.

dotemacs: Yes, I agree.

PR #434, opened 2015-11-10
Remove build controller

novotnyjakub:
This PR moves the last remaining bit to the ProcessController and removes wicked + BuildController altogether. 

I had to upgrade `capybara-webkit` to 1.7 and it now raises a deprecation warning if compiled with Qt4, so I recommend to compiling it with Qt5 (a qt5 package with homebrew). 

We should bare in mind that the ProcessController now is quite heavy and even has the Rubocop warning for ClassLength supressed. So the next step is to think about what the structure should be and start improving it. But at least we have complete control over the process.


colinbruce:
Wow! :+1:

spec/support/feature_helpers.rb
```diff
@@ -0,0 +1,10 @@
+module FeatureHelpers
+  def start_new_application
+    visit '/'
+    click_link 'Start now'
+  end
+end
+
+RSpec.configure do |c|
+  c.include FeatureHelpers, type: :feature
+end
```
dotemacs: What do we gain with this change?

Why didn't **visit applications_new_path** work?

dotemacs:
What @colinbruce said.

spec/support/feature_helpers.rb
novotnyjakub: `applications_new_path` isn't a thing any more. It used to be a GET route, which created a new application and redirected to the build. I've made the initial button a POST, which routes to `/applications` and that creates + redirects. 

That means we're not creating objects in a GET action any more, which is correct from RESTful perspective. But we therefore can't simply visit a page to create a new `Application`.

dotemacs: Ok, I'll have to look at this properly tomorrow

dotemacs:
Merged master into this, but didn't want to run:

```
git checkout master
git merge --no-ff remove-build-controller
git push origin master
```

just yet, as I wanted to see how it plays out on TravisCI.

dotemacs:
@novotnyjakub ^

novotnyjakub:
Cheers! Just ran through it quickly and it looks good, so if CI passes, we should be good to merge.

dotemacs:
@novotnyjakub we should update **.travis.yml** with the latest qt5 package on debian/ubuntu, so that we suppress the following errors:

```
...WARNING: The next major version of capybara-webkit will require at least version 5.0 of Qt. You're using version 4.8.1.
...
```

dotemacs:
@novotnyjakub thanks, just please double check 6e8b784

novotnyjakub:
Looks good to me.

dotemacs:
GREEN!

dotemacs:
Cool, will merge.

PR #437, opened 2015-11-10
Renamed Payment to PartPayment

dotemacs:
This was done so that it fits with the real world object/action it's
trying to model.


app/models/query/processed_applications.rb
```diff
@@ -6,20 +6,20 @@ def initialize(user)
...
 
     def where_condition
-      <<WHERE
-(applications.application_outcome IS NOT NULL)
-AND
-(evidence_checks.id IS NULL OR evidence_checks.completed_at IS NOT NULL)
-AND
-(payments.id IS NULL OR payments.completed_at IS NOT NULL)
-WHERE
+      <<-WHERE.gsub(/^\s+\|/, '')
```
novotnyjakub: Isn't this unnecessary gsub on every sql query? I can see that it fixes indentation, but I think it's not worth doing.

spec/models/query/waiting_for_part_payment_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe Query::WaitingForPartPayment, type: :model do
+  describe '#find' do
+
+    let(:user) { create :user }
+    let!(:application1) { create :application, user_id: user.id, office_id: user.office_id }
+    let!(:application2) { create :application, user_id: user.id, office_id: user.office_id }
+    let!(:application3) { create :application, user_id: user.id, office_id: user.office_id }
+    let!(:payment1) { create :part_payment, application: application1, expires_at: 2.days.from_now }
```
novotnyjakub: These are not converted, but they're not actually being used as variables in the spec. We could just convert them to `before` block.

app/models/query/processed_applications.rb
dotemacs: Ruby style guide: https://github.com/bbatsov/ruby-style-guide#heredocs

And if this is going to slow the app tests down, in that case I think we have bigger problems elsewhere.

spec/models/query/waiting_for_part_payment_spec.rb
dotemacs: End of the day, slow... Please add more context/example :)

app/models/query/processed_applications.rb
novotnyjakub: Cool, didn't know that was a recommended style :+1: 

I didn't mean running tests, though, but the real app. This runs every time somebody access `/processed_application`. But yeah, not a biggie, I guess, I'm probably overthinking it.

spec/models/query/waiting_for_part_payment_spec.rb
novotnyjakub: :-) 

I was thinking: 

```ruby
let(:application1) { create :application, user_id: user.id, office_id: user.office_id }
let(:application2) { create :application, user_id: user.id, office_id: user.office_id }
let(:application3) { create :application, user_id: user.id, office_id: user.office_id }

before do
  create :part_payment, application: application1, expires_at: 2.days.from_now 
  create :part_payment, application: application2, expires_at: 1.days.from_now 
  create :part_payment, application: application3, expires_at: 1.days.from_now, completed_at: 2.days.ago 
end
```

Which would remove the bangs as well and made clear that the payments are not being used for the test results.

novotnyjakub: But don't worry about it too much, it's a simple test..

dotemacs: Cool, thanks for the comments, will pick it up tomorrow!

dotemacs: Can't be done because:

```
Failure/Error: let(:payment1) { create :part_payment, application: application1, expires_at: 2.days.from_now }
       
`let` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```

dotemacs:
This change is now ready to go. All the tests pass.

@novotnyjakub & @colinbruce can you review it when you can? 

Thanks

spec/models/query/waiting_for_part_payment_spec.rb
novotnyjakub: There are no `let` in the `before` block ;-) They're not needed, because the payment variables are not asserted against anything in the test.

dotemacs: Cool, re-implementing it now...

dotemacs:
Re-implemented the example suggested.

@novotnyjakub & @colinbruce can you review it when you can?

Thanks

novotnyjakub:
:+1: all good, merge at will!

PR #441, opened 2015-11-12
Rename application outcome to outcome

dotemacs:
No need for the **`application_'** part of the name, especially since this is a column in **`applications'** table.

Also added a **'missed'** migration for a foreign key for applications on benefit_override table.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151110090803) do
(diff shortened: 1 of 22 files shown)


db/schema.rb
```diff
@@ -208,6 +208,7 @@
   add_foreign_key "applications", "users"
   add_foreign_key "benefit_checks", "applications"
   add_foreign_key "benefit_checks", "users"
+  add_foreign_key "benefit_overrides", "applications"
```
colinbruce: Is this still sneaking in and out? :)

colinbruce:
Looks fine, merge on green!

PR #442, opened 2015-11-12
Extend resolver service

colinbruce:
This ensures that all three models: Application, EvidenceCheck and PartPayment use the same method of setting their completed data fields.

Next step: 
Move the creation of EvidenceChecks and PartPayments into the ResolverService
This will reduce the work done in the controllers


dotemacs:
Looks good

PR #446, opened 2015-11-13
Benefit override outcome recording

dotemacs:
When the benefit override flow completes, it should record the outcome of the benefit override in the application itself.

It previously wasn't doing this, but with this PR it will.

The reason I picked this up is because I started work on a feature which would allow the users to remove the applications which are waiting for on the dashboard.


colinbruce:
This looks good. As discussed, we should probably add a task to update existing records without an outcome set.

colinbruce:
Nice, go for merge once green! :+1:

PR #448, opened 2015-11-16
Add explicit tests for refund dates

colinbruce:
Also, refactored some of the tests to use the
subject-result constructs.


dotemacs:
Looks OK

PR #450, opened 2015-11-17
Updated processed applications

colinbruce:
* updated query to only include applications where completed_at is set
* extended tests
* created appropriate factory


dotemacs:
Great, merge away!

PR #451, opened 2015-11-17
Added authentication for Evidence controller

dotemacs:
Because only the authenticated users should have access to this
controller and its actions.


colinbruce:
Looks fine but, as discussed, setting up shared_examples (or similar) makes more sense long term.

:+1:  Merge on green

PR #452, opened 2015-11-17
Removed 'focus' tags from specs

dotemacs:
No need for them to be left lying around, so removing them.


colinbruce:
Looks fine to me, but it might be worth waiting for @novotnyjakub to explain why he added them in the first place!

PR #453, opened 2015-11-17
Added validation for income

colinbruce:
The Income form needed validation on the income field.

The page could not be saved without it and was then throwing a 500 error
if the page was saved without an income recorded.


dotemacs:
Merge it

PR #455, opened 2015-11-17
updated content for return evidence

leighmoney:


dotemacs:
Will merge this in, as it'll be merged into 'return-application' branch which will be re-tested. So I'll fix any issues on that branch if there are any.

PR #457, opened 2015-11-17
Updated log_stuff

colinbruce:
* refactored the log method to reduce the complexity  
* spread the single method into three parts


dotemacs:
Good to merge

PR #464, opened 2015-11-19
Extend return application

colinbruce:
@dotemacs optional PR that moves that reuses the ResolverService and moves the setting
of the application into it too


colinbruce:
Example only - not needed

PR #465, opened 2015-11-19
Add decision fields to application

colinbruce:
This will allow new code to set the values.
Meanwhile, a separate PR will be added to
migrate the data for existing, complete, records.


dotemacs:
good to go

PR #466, opened 2015-11-19
Extended ResolverService

dotemacs:
Enabled it to resolve the decision of the application on the application
itself, but also to update the corresponding object by which the
application decision was derived.


colinbruce:
Looks great! :+1:  Merge on :green_heart:

PR #467, opened 2015-11-20
Upgraded RSpec to 3.4

dotemacs:
Upgraded as the dependancies should not lag behind too much behind the
latest stable releases

The errors needed to be specified explicitly for the warning messages to
be suppressed.


colinbruce:
Looks good, merge on green.

PR #469, opened 2015-11-20
Fix for running db:migrate from scratch

colinbruce:
Because the DwpCheck model had been deleted, this would cause db:migrate to fail
when run from scratch.

Switched the script to execute SQL rather the use the ActiveRecord::connection method.


dotemacs:
Looks fine

PR #470, opened 2015-11-20
Return part payment applications

dotemacs:
Return PartPayments using the ResolveService

### DIFF
--- a/app/controllers/part_payments_controller.rb
+++ b/app/controllers/part_payments_controller.rb
@@ -27,7 +27,7 @@ class PartPaymentsController < ApplicationController
   end
 
   def summary_save
-    ResolverService.new(part_payment, current_user).resolve('return')
+    ResolverService.new(part_payment, current_user).process
     redirect_to(confirmation_part_payment_path(part_payment))
   end
 
--- a/app/controllers/part_payments_controller.rb
+++ b/app/controllers/part_payments_controller.rb
@@ -36,6 +36,14 @@ class PartPaymentsController < ApplicationController
     @result = Views::PartPayment::Result.new(part_payment)
   end
 
+  def return_letter
+    application_overview
+  end
+
+  def return_application
+    redirect_to root_path if ResolverService.new(part_payment, current_user).resolve('return')
+  end
+
   private
 
   def part_payment
@@ -44,6 +52,10 @@ class PartPaymentsController < ApplicationController
 
   def application
     part_payment.application
+  end
+
+  def application_overview
+    @overview = Views::ApplicationOverview.new(part_payment.application)
   end
 
   def accuracy_params
--- a/app/controllers/part_payments_controller.rb
+++ b/app/controllers/part_payments_controller.rb
@@ -41,7 +41,12 @@ class PartPaymentsController < ApplicationController
   end
 
   def return_application
-    redirect_to root_path if ResolverService.new(part_payment, current_user).resolve('return')
+    if ResolverService.new(part_payment, current_user).resolve('return')
+      redirect_to root_path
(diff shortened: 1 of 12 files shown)


colinbruce:
@dotemacs The tests are green, but I've just spotted an issue.  I'll pick this up next week as you are on IRAT.

app/controllers/part_payments_controller.rb
```diff
@@ -36,6 +36,14 @@ def confirmation
     @result = Views::PartPayment::Result.new(part_payment)
   end
 
+  def return_letter
+    application_overview
+  end
+
+  def return_application
+    redirect_to root_path if ResolverService.new(part_payment, current_user).resolve('return')
```
novotnyjakub: If the `ResolverService` fails to resolve, this will raise an error because the template is missing.

spec/features/return_applications/return_part_payment_applications_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    context 'processing an application' do
+      before { click_link application1.reference }
+      scenario 'shows the application data' do
+        expect(page).to have_content 'Process part-payment'
+        expect(page).to have_content application1.applicant.full_name
+      end
+
+      scenario 'shows "Start now" button' do
```
novotnyjakub: I would join this scenario with the previous one and the next one as well. I know that generally it's a good practice to have only one assertion per text, but all these only assert the page content. There's no action which would make the page different and running 3 scenarios is unnecessarily slow, I think.

spec/features/return_applications/return_part_payment_applications_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    scenario 'when returning application' do
+      click_link application1.reference
+      click_link 'Return application'
+      expect(page).to have_content 'Processing complete'
+      expect(page).to have_button 'Finish'
+      click_button 'Finish'
+      expect(page).to have_content 'Start now'
+      within '.waiting-for-part_payment' do
+        expect(page).not_to have_content(application1.reference)
```
novotnyjakub: `not_to` doesn't reliably work with `have_content`. There's a dedicated matcher `have_no_content` for this in Capybara.

colinbruce:
@novotnyjakub can you merge this if you are happy? Thanks

app/services/resolver_service.rb
```diff
@@ -28,18 +43,14 @@ def mark_complete
...
-      'return' => 'none' }
-  end
-
   def derive_object
     @calling_object.class.name.underscore
   end
 
   def lookup_decision(outcome)
-    send(derive_object)[outcome]
+    { 'full' => 'full',
```
dotemacs: There's a fifth option. Please don't merge this yet, thanks

novotnyjakub:
Ok, so we ready to merge? I'm a bit confused with this now :) The code looks good to me, but I don't understand the "No tests!" comment, should we wait?

dotemacs:
Open the specs for ResolverService, do you see any tests that are testing for 'undetermined'?

This is what the 'No tests!' comment refers to. Also, that comment was made on a specific commit, rather than on the whole PR.

colinbruce:
Proposal: I rollback the last('Added "undetermined" to lookup') commit.
This is currently not called, in-fact the *only* value passed in currently by the code is hard coded to be 'return'.
When we extend the resolver status to be used by all valid types, we add the undetermined value to the lookup along with the requisite tests!

novotnyjakub:
I like that, always better to add things when needed, I think.

PR #476, opened 2015-11-25
Validate refund date relative to date received

novotnyjakub:
This changes validation of the refund date form field. It now has to be:
* up to 3 months before the date application received
* not later than date application received


colinbruce:
Looks good: :+1:

PR #479, opened 2015-11-25
Fix missed translations after form renaming

novotnyjakub:


colinbruce:
Looks good, merge on green!

PR #485, opened 2015-11-28
Front-end changes

colinbruce:
Consolidated front-end changes
* Added the date hint text
* Deleted the Hint text spans from the page
* Updated the yml file to add _html suffix to the optional
label, this removes need from putting html safe on view
* removed span.hint css
* updated the date hint text to point at the correct label

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -224,6 +224,3 @@ input.medium-field{
     width:70%;
 }
 
-span.hint {
-  padding-left:5px;
-}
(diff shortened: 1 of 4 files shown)


colinbruce:
@kelliematheson /@leighmoney With the [hint text ticket](https://www.pivotaltracker.com/story/show/106632440), was there a particular date we should show?

kelliematheson:
@colinbruce The copy to use is in the ticket already, the grey hint text should have this exact whole line of copy:

For example DD/MM/YYYY

So it will look like this:

Date of birth
For example DD/MM/YYYY
[                                  ]

colinbruce:
@kelliematheson Ready for testing on https://front-end-combined.helpwithcourtfees.dsd.io/

kelliematheson:
Tested on branch builder and approve

PR #486, opened 2015-11-30
Use date processed on waiting applications list

novotnyjakub:
Instead of displaying `Expires at` we now display `Date processed`. Also the list is ordered by the processed date of the application from eldest to newest.


colinbruce:
LGTM, merge on green :+1:

PR #488, opened 2015-12-01
Return nil when completed_at is missing

novotnyjakub:
It was causing an error on applications which were not completed.


colinbruce:
Go, go, go... :+1:

PR #490, opened 2015-12-01
Moved manager links to top bar

colinbruce:
* removed from the manager's home page
* removed home page tests

[finishes #105864428]

@kelliematheson [ticket](https://www.pivotaltracker.com/story/show/105864428)


kelliematheson:
Tested on branchbuilder, accepted

PR #494, opened 2015-12-01
State machine

novotnyjakub:
Introducing a simple state to the `Application` model to keep track of the overall journey. 

No special gem is used, just ActiveRecord `enum`, which provides the states on `Application` we need. The state transitions themselves should be concentrated in service classes (ie. `ResolverService`). 

The current states are: 

* created
* evidence_check
* part_payment
* processed

This PR centralises completion of `Application`, `EvidenceCheck` and `PartPayment` to the `ResolverService`. This service class is now solely responsible for completing those respective objects, creating `EvidenceCheck`/`PartPayment` as needed and also setting `decision` and `decision_type` on `Application` when moving to the final `processed` state.

I think we should also consider removing the query objects are with the state scopes, they may not be needed.

We should also probably remove `EvidenceCheck` -> `Evidence`, because that's what we use in the UI.


novotnyjakub:
@colinbruce and @dotemacs please review carefully including manual testing.

colinbruce:
Reviewed and happy to merge. :+1:

PR #496, opened 2015-12-04
Updated management views

colinbruce:
Following a CSS update the management views for office and staff
had become... sub-optimal
screen shot 2015-12-04 at 11 20 14screen shot 2015-12-04 at 11 20 27

This update corrects the spacing and adds a row for the BE code.

This means that users will have a better experience on the management side of the app, until the 'merging staff and office' ticket reaches the backlog.


kelliematheson:
Looks goods.

PR #500, opened 2015-12-08
Deleted applications

novotnyjakub:
User can mark application as deleted from the detail of processed application. They have to provide a reason for the deletion. 

Deleted applications are then displayed in the new section, which has almost identical behaviour to processed applications.


colinbruce:
Do we need the app/controllers/concerns/.keep file?

novotnyjakub:
No, that's why I removed it ;-)

novotnyjakub:
I've implemented the missing bit which was adding audit data `deleted_at` and `deleted_by`. The problem with the solution is that if we carry on with the approach we'd have a pair of columns for each state, to track who and when did it. Maybe we should consider using another solution for this, but for now to move on, it's implemented like this.

novotnyjakub:
With the codeclimate check, I don't think I can do something about it now. The reason for this is that we're delegating a lot of methods to the split models. That work has never been fully completed so until that's done, I guess we'll have to live with it.

This issue doesn't actually decrease the GPA, so I wouldn't worry about it too much.

colinbruce:
Agreed on the codeclimate point. 

LGTM, merge when ready. :+1:

PR #501, opened 2015-12-09
Fix for entity code test fails

colinbruce:
## Updated the office factory 
- The factory generates sequences of entity_codes
- Using ljust creates 100, 200, 300, 400...
- Switching to rjust means that 001, 002 sequences are created, this
leads to less opportunities to clash.

### DIFF
--- a/spec/factories/offices.rb
+++ b/spec/factories/offices.rb
@@ -5,7 +5,7 @@ FactoryGirl.define do
     end
 
     sequence(:name) { |n| "#{Faker::Company.name} #{n}" }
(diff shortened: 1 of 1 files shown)


novotnyjakub:
It's fine by me, but I don't think it'll make a difference. I'm afraid the clashes occur because of some weird rspec concurency.

colinbruce:
I ran the ./spec/services/evidence_check_selector_spec.rb test 100 times with ljust and rjust

7% of the ljust runs failed, 0% of the rjust ones did!

PR #503, opened 2015-12-10
Reference number after 1.1.2016.

dotemacs:
The change is to show the reference number next to the application after 1.1.2016. so that we can do away with the Excel based Remission Register.

This PR implements this ticket:

https://www.pivotaltracker.com/n/projects/1212124/stories/109520848

### DIFF
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -1,7 +1,7 @@
 module ReferenceHelper
 
   def after_desired_date?
-    Time.zone.today > Date.parse('2015-12-31')
+    Time.zone.today > Settings.reference.date
   end
 
   def table_header
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -16,7 +16,7 @@ module ReferenceHelper
   end
 
   def table_column(application)
-    prefix = "<td> <a href=\"/processed_applications/#{application.id}\">"
+    prefix = "<td> <a href=\"/#{controller.controller_name}/#{application.id}\">"
     if after_desired_date?
       "#{prefix}#{application.reference}</a> </td>\n<td> #{application.applicant} </td>".html_safe
     else
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -31,4 +31,8 @@ module ReferenceHelper
       %w[processed_on processed_by]
     end
   end
+
+  def display_reference(application)
+    after_desired_date? ? application.reference : ''
+  end
 end
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -7,7 +7,7 @@ module ReferenceHelper
   def table_header
     applicant = t('processed_applications.table_header.applicant')
 
-    if Time.zone.today > Date.parse('2015-12-31')
+    if after_desired_date?
       reference = t('processed_applications.table_header.reference')
       "<th> #{reference} </th>\n<th> #{applicant} </th>".html_safe
     else
@@ -17,7 +17,7 @@ module ReferenceHelper
 
   def table_column(application)
     prefix = "<td> <a href=\"/processed_applications/#{application.id}\">"
-    if Time.zone.today > Date.parse('2015-12-31')
+    if after_desired_date?
       "#{prefix}#{application.reference}</a> </td>\n<td> #{application.applicant} </td>".html_safe
     else
       "#{prefix}#{application.applicant}</a> </td>".html_safe
@@ -25,7 +25,7 @@ module ReferenceHelper
   end
 
   def processing_details_options
-    if Time.zone.today > Date.parse('2015-12-31')
+    if after_desired_date?
       %w[processed_on processed_by reference]
     else
       %w[processed_on processed_by]
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -1,5 +1,9 @@
 module ReferenceHelper
 
+  def after_desired_date?
+    Time.zone.today > Date.parse('2015-12-31')
+  end
+
   def table_header
     applicant = t('processed_applications.table_header.applicant')
 
--- a/app/helpers/reference_helper.rb
+++ b/app/helpers/reference_helper.rb
@@ -1,4 +1,4 @@
-module ReferenceTableHelper
+module ReferenceHelper
 
   def table_header
     applicant = t('processed_applications.table_header.applicant')
--- a/app/views/applications/process/confirmation.html.slim
+++ b/app/views/applications/process/confirmation.html.slim
@@ -3,6 +3,8 @@ header
     .small-12.medium-12.large-12.columns
       h2 Processing complete
 
+= display_reference(@application)
+
 .row
   .small-12.medium-8.large-5.columns
     = build_section 'Results', @confirm, @confirm.all_fields
--- a/app/views/shared/processed_and_deleted/_index.html.slim
+++ b/app/views/shared/processed_and_deleted/_index.html.slim
@@ -8,7 +8,7 @@
       table *{class: "#{state}-applications"}
         thead
           tr
-            th =t('processed_applications.table_header.applicant')
+            = table_header
             th =t('processed_applications.table_header.date_received')
             th =t('processed_applications.table_header.processed_on')
             th =t('processed_applications.table_header.processed_by')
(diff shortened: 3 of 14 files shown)


app/helpers/reference_helper.rb
```diff
@@ -0,0 +1,38 @@
+module ReferenceHelper
+
+  def after_desired_date?
```
dotemacs: I'm not quiet sure about this method, but I needed to add it as repeating the method's logic all over the place seemed wrong.

app/views/shared/processed_and_deleted/_index.html.slim
```diff
@@ -8,7 +8,7 @@
       table *{class: "#{state}-applications"}
         thead
           tr
-            th =t('processed_applications.table_header.applicant')
+            = table_header
```
colinbruce: Shouldn't this be replicated in the data below?  You are changing the heading to say reference, but the link always displays applicant

colinbruce: Can this be called explicitly? i.e. ReferenceHelper.table_header, it's not very clear where it points otherwise.

dotemacs: About the table column heading & rows, good catch. I was working on a outdated copy of the app, updating it now!

dotemacs: As for using the helper:

```ruby
ReferenceHelper.table_header
```

it throws an error using it that way also all the other helpers are used without specifying the module name.

app/helpers/reference_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    Time.zone.today > Date.parse('2015-12-31')
```
novotnyjakub: Can we put this to `settings.yml` and read it from the `Settings` class? Maybe the `settings.yml` should even read the value from `ENV` so that we don't have to do a full release if this changes.

app/helpers/reference_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    if after_desired_date?
+      reference = t('processed_applications.table_header.reference')
+      "<th> #{reference} </th>\n<th> #{applicant} </th>".html_safe
+    else
+      "<th> #{applicant} </th>".html_safe
+    end
+  end
+
+  def table_column(application)
+    prefix = "<td> <a href=\"/#{controller.controller_name}/#{application.id}\">"
```
novotnyjakub: I don't think this is safe, can we use url helpers? You can use `send` and generate its name dynamically as well.

spec/helpers/reference_helper_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'rails_helper'
+
+RSpec.describe ReferenceHelper, type: :helper do
+
+  describe '#after_desired_date?' do
+    context 'when the date is before 1st January 2016' do
+      it 'returns false' do
+        Timecop.freeze(Date.new(2015, 12, 31)) do
+          expect(helper.after_desired_date?).to eq(false)
```
novotnyjakub: Just a little one, we should be using `be(false)`.

app/views/applications/process/confirmation.html.slim
```diff
@@ -3,6 +3,8 @@ header
     .small-12.medium-12.large-12.columns
       h2 Processing complete
 
+= display_reference(@application)
```
novotnyjakub: This will have no formatting, will it? Should we have the same visual style as on the evidence check letter when application is selected for evidence?

novotnyjakub:
I can now see why this is quite annoying and making the codebase dirty. Can we please have a card to clean it up after the launch date? Just so we don't forget.

app/views/applications/process/confirmation.html.slim
dotemacs: @novotnyjakub 
> This will have no formatting, will it?

Correct, discussed it with Kellie already and I mention it in the commit message: https://github.com/ministryofjustice/fr-staffapp/commit/028faf9432d0975864cefcd754d57555ba4d35f1

Once we know what the style should be applied, it'll be added.

spec/helpers/reference_helper_spec.rb
dotemacs: Adding it now :dash:

app/helpers/reference_helper.rb
dotemacs: @novotnyjakub 

> Can we put this to settings.yml and read it from the Settings class? Maybe the settings.yml should even read the value from ENV so that we don't have to do a full release if this changes.

We know where this one value resides, in this helper. The amount of time and seeing how we already have a ticket to delete this feature after 1st January: https://www.pivotaltracker.com/story/show/109815848 I'd argue that this would not be worth it, even though from the design perspective your suggestion is cleaner, but only if this was going to be a feature that would stay. Seeing as it's only temporary, I'd say to leave it as is.

app/helpers/reference_helper.rb
dotemacs: I'm all for using more secure implementation, but I'm not sure exactly what you're suggesting. Can you give me a hint?

Also, just to point out that this helper will be used by:

* deleted_applications#show
* processed_applications#show

paths so it needs to be 'generic'.

dotemacs:
@novotnyjakub 
> I can now see why this is quite annoying and making the codebase dirty. Can we please have a card to clean it up after the launch date? Just so we don't forget.

Already done: https://www.pivotaltracker.com/story/show/109815848 I asked Adrian to do this yesterday.

app/helpers/reference_helper.rb
novotnyjakub: :+1: you're right, let's not overdo it.

app/helpers/reference_helper.rb
novotnyjakub: You should be able to use: 

```ruby
url_for(controller: controller.controller_name, action: show, id: application.id)

# or maybe even just

url_for(action: show, id: application.id)
```

app/views/applications/process/confirmation.html.slim
novotnyjakub: :+1:

app/helpers/reference_helper.rb
novotnyjakub: But actually, there's quite a big reason for ENV, which is testability. We need to be able to deploy to demo (or somewhere) and switch it on now.

dotemacs: What you say is right, testability. So we have two approaches:

* use ENV var as you suggest
* get Filippos to fast forward the time on the demo server by 1 month and it'll be testable (then reset the server's time back to the correct time)

With both approaches we'll require Filippos' time. With the first approach, it'll be setting the ENV var in the **fr-deploy** repo plus me modifying the code. In the second, as mentioned above.

The question is, what trade off do we want to make: doing something cleanly that'll take a bit more time or going with what we've got that'll take less time?

app/helpers/reference_helper.rb
dotemacs: In theory this method can be rewritten as:

```ruby
  def table_column(application)
    if after_desired_date?
      "<td> #{link_to(application.reference, url_for(controller: controller.controller_name, action: :show, id: application.id))} </td>\n<td> #{application.applicant} </td>".html_safe
    else
      "<td> #{link_to(application.applicant, url_for(controller: controller.controller_name, action: :show, id: application.id))} </td>".html_safe
    end
  end
```

Which looks cleaner.

Whereas my solution, aesthetically unpleasant, it's already functional with the tests passing. 
(And we'll be scrapping the whole thing in three weeks.)

But then we introduce the following:

* the tests now complain about: 

```
     ActionController::UrlGenerationError:
       No route matches {:action=>"show", :controller=>"test", :id=>1}
```
* what do I do about the line lengths for RuboCop?

Also, how is my solution insecure? Just curious, because the controller and the application argument is something the app will pass to the helper, they won't be something obtained from the user.

novotnyjakub: Not insecure, but not safe in terms of result, because routes != controller names. But yeah, if we make sure it's gone in 3 weeks, I'm not gonna argue.

app/helpers/reference_helper.rb
novotnyjakub: We can set ENVs ourselves, without anyone's involvement. What I think would be good is to read ENV, but fallback to 1.1.2016. So like that, we don't even have to high state production environments when we deploy. We'd just set it on demo and branch builder will automatically pick it up.

Moving time on a server would be more work I think, everybody would need to be ready at the same time and there's risk involved. Also some data would be created in the future and then the time would move back..

novotnyjakub:
OK, lgtm now :+1: Not sure what the testing plan is..

PR #505, opened 2015-12-11
Allow the date fee paid to match received date

dotemacs:
Allow the date fee paid to be on the same day as the date the
application was received.

Previously it only allowed the paid fee date to be the date before.

:ticket::
https://www.pivotaltracker.com/story/show/109797244


novotnyjakub:
:+1:

PR #507, opened 2015-12-14
Business entities

novotnyjakub:
This introduces `BusinessEntity` model which will be managed by HQ. `BusinessEntity` has to exist for each combination of `Office` and `Jurisdiction` in use. Editing of business entities will be added later.

A big change is that `reference` for `Application` is now only created on completion of application. Until then it's `NULL`. Because the prefix for the reference is now taken from the relevant `BusinessEntity`, reference to it is also stored to `Application`.

### DIFF
--- a/app/builders/application_builder.rb
+++ b/app/builders/application_builder.rb
@@ -10,22 +10,12 @@ class ApplicationBuilder
     Application.create(
       office_id: @user.office_id,
       user_id: @user.id,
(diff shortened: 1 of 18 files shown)


dotemacs:
Looks fine to me

colinbruce:
Looks good.
Problem offices on Prod:
14 - To be deleted
4 and 7, need BEcodes adding before migration can run.

PR #508, opened 2015-12-16
Lock down invites for managers

dotemacs:
This PR fixes this privilege escalation issue: https://www.pivotaltracker.com/story/show/102069424

Don't allow the managers to invite admins.

In fact this was not allowed, but the PR prevents even the malicious attempts at privileged invitation attempt.

I'm aware that some of the controller tests can be interpreted as feature tests, but since this is a security related, I'm happy to violate test purity for the sake of security.

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -38,6 +38,6 @@ class Users::InvitationsController < Devise::InvitationsController
   end
 
   def user_not_admin_and_role_is_admin?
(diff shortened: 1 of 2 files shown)


novotnyjakub:
Looks alright. 

What about regular users, though, should they be able to invite other users?

dotemacs:
The regular users can't invite anybody, so didn't bother testing for them.

In fact, I was being pedantic when adding the tests for admin users, as this was only affecting manager users.

colinbruce:
Unpriviledged should beUnprivileged, but otherwise, looks good

PR #509, opened 2015-12-16
Add more decision columns

novotnyjakub:
This adds `decision_date` and `decision_cost` to `Application`. The main purpose is to be able to easily access the summary of decided application. After this we may consider taking all the `decision_` columns to a separate table, which then can be used for various stats (like finance).

It also migrates all existing processed applications to have all the decision columns filled in (including `decision` and `decision_type`.

### DIFF
--- a/app/services/resolver_cost_calculator.rb
+++ b/app/services/resolver_cost_calculator.rb
@@ -1,35 +1,32 @@
 class ResolverCostCalculator
   def initialize(source)
     @source = source
+    @application = @source.is_a?(Application) ? @source : @source.application
   end
 
   def cost
-    @source.outcome == 'none' ? 0 : incurred_cost(@source)
+    @source.outcome == 'none' ? 0 : incurred_cost
   end
 
   private
 
-  def incurred_cost(source)
-    if source.is_a?(PartPayment)
-      if source.application.evidence_check.present?
-        fee(source) - amount_to_pay(source.application.evidence_check)
-      else
-        fee(source) - amount_to_pay(source.application)
-      end
+  def incurred_cost
+    if @source.is_a?(PartPayment)
+      fee - amount_to_pay
     else
-      fee(source)
+      fee
     end
   end
 
-  def fee(source)
-    if source.is_a?(Application)
-      source.detail.fee
-    else
-      source.application.detail.fee
-    end
+  def fee
+    @application.detail.fee
   end
 
-  def amount_to_pay(source)
-    source.try(:amount_to_pay) || 0
+  def amount_to_pay
+    if @application.evidence_check.present?
+      @application.evidence_check.try(:amount_to_pay) || 0
+    else
+      @application.try(:amount_to_pay) || 0
+    end
   end
 end
--- a/app/services/resolver_cost_calculator.rb
+++ b/app/services/resolver_cost_calculator.rb
@@ -1,0 +1,35 @@
+class ResolverCostCalculator
+  def initialize(source)
+    @source = source
+  end
+
+  def cost
+    @source.outcome == 'none' ? 0 : incurred_cost(@source)
+  end
+
+  private
+
+  def incurred_cost(source)
+    if source.is_a?(PartPayment)
+      if source.application.evidence_check.present?
+        fee(source) - amount_to_pay(source.application.evidence_check)
+      else
+        fee(source) - amount_to_pay(source.application)
+      end
+    else
+      fee(source)
+    end
+  end
+
+  def fee(source)
+    if source.is_a?(Application)
+      source.detail.fee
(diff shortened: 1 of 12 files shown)


spec/lib/decision_date_and_cost_migration_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    create(:application_no_remission, :processed_state, fee: 500, decision: 'none', decision_date: nil, decision_cost: nil)
+  end
+
+  describe '#run!' do
+    before do
+      migration.run!
+    end
+
+    it 'sets the correct decision_date and decision_cost' do
+      # for processed application without evidence check or part payment
```
dotemacs: Not sure what's the best alternative, but just the fact that there are so many test comments tells me that there should be another way to document this. Possibly breaking it up into smaller examples ?

novotnyjakub: So the reason why it's in a single test is that I need to make sure that the migration runs on the whole DB at the same time, but each individual scenario is valid. If I tested them in isolation, it wouldn't be guaranteed.

I totally agree though that it's not ideal. On the other hand all the 3 migration classes will be deleted once run and verified.

app/services/resolver_cost_calculator.rb
```diff
@@ -0,0 +1,35 @@
...
+    if source.is_a?(PartPayment)
+      if source.application.evidence_check.present?
+        fee(source) - amount_to_pay(source.application.evidence_check)
+      else
+        fee(source) - amount_to_pay(source.application)
+      end
+    else
+      fee(source)
+    end
+  end
```
dotemacs: Maybe instead of `incurred_cost` as above with all the nested logic implement something like this:

```ruby
def incurred_cost(source)
  process_part_payment
end

def process_part_payment
  return process_evidence_check if source.is_a?(PartPayment)
  fee(source)
end

def process_evidence_check
  another_amount = if source.application.evidence_check.present?
                     source.application.evidence_check
                   else
                     source.application
                   end
  work_out_the_sum(another_amount)
end

def work_out_the_sum(another_amount)
  fee(@source) - amount_to_pay(another_amount)
end
```

or maybe even:

```ruby

def incurred_cost
  process_part_payment
end

def process_part_payment
  return work_out_the_sum if @source.is_a?(PartPayment)
  fee(source)
end

def process_evidence_check
  if @source.application.evidence_check.present?
    @source.application.evidence_check
  else
    @source.application
  end
end

def work_out_the_sum
  fee(@source) - amount_to_pay(process_evidence_check)
end
```

dotemacs: I think the names of the methods can be changed to be a bit more descriptive. 

My main point was to avoid the nested if-ing.

spec/lib/decision_date_and_cost_migration_spec.rb
dotemacs: Ah, OK, in that case run with it. Thanks for the explanation!

app/services/resolver_cost_calculator.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def fee(source)
+    if source.is_a?(Application)
+      source.detail.fee
+    else
+      source.application.detail.fee
+    end
+  end
```
dotemacs: The same thing here, maybe it can be re-worked so that `source` is not passed around, but the instance variable used instead?

app/services/resolver_cost_calculator.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def amount_to_pay(source)
+    source.try(:amount_to_pay) || 0
+  end
```
dotemacs: And again here, maybe use the instance instead of passing the argument in...

app/services/resolver_cost_calculator.rb
novotnyjakub: I'll have a look what I can do. This suggestion I find personally more confusing, because for example `incurred_cost != process_part_payment` which this method suggest. But I'll see what I can do tomorrow morning with fresh eyes.

app/services/resolver_cost_calculator.rb
novotnyjakub: It's not the same thing, actually. But yeah, I agree that the naming is not ideal and I'll redo it a bit.

novotnyjakub:
@dotemacs I've now refactored the `ResolverCostCalculator`, please have a look.

dotemacs:
@novotnyjakub I like what you've done with `ResolverCostCalculator`, much cleaner :) thanks

colinbruce:
LGTM :+1:

PR #513, opened 2015-12-18
Error message for priviledge escalation

dotemacs:
Provided a meaningful error message to the users when they try to invite a admin, but they themselves are a manager.


colinbruce:
Looks good, works well. Merge at will!

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -38,9 +40,10 @@
       end
 
       it 'does not allow you to invite admins as a manager' do
-        expect {
-          post :create, user: admin_invitation
-        }.to raise_error 'Unprivileged invitation error: Non-admin user is inviting an admin.'
+        post :create, user: admin_invitation
+        expect(response).to redirect_to(new_user_invitation_path)
+        get :new
+        expect(response.body).to include 'You cannot create an admin account'
```
colinbruce: You can test explicitly for flash messages like this...
https://github.com/ministryofjustice/fr-staffapp/blob/a7dc9217dcd2ce4fdc1666193b3a5bc0bf201112/spec/controllers/users_controller/admin_user_spec.rb#L189

dotemacs: You are right. I figured that if the error message content is present, then... Thank you

PR #515, opened 2015-12-18
Finance report

colinbruce:
Allow admin users to output the finance data from between two dates to CSV

After login this will be accessible at outputs/finance_report

[Ticket](https://www.pivotaltracker.com/story/show/93418134)

[finishes #93418134]

### DIFF
--- a/app/controllers/outputs_controller.rb
+++ b/app/controllers/outputs_controller.rb
@@ -12,13 +12,23 @@ class OutputsController < ApplicationController
 
   def finance_report_generator
     authorize! :access, :outputs
-    @date_from = Date.parse(report_params[:date_from])
-    @date_to = Date.parse(report_params[:date_to])
-    @data = FinanceReportBuilder.new(report_params[:date_from], report_params[:date_to])
-    send_data @data.to_csv, filename: "finance-report-#{@date_from}-#{@date_to}.csv"
+    @form = form
+    if @form.valid?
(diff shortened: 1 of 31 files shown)


spec/models/forms/finance_report_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it { is_expected.to validate_presence_of(:office) }
+    it { is_expected.to validate_presence_of(:jurisdiction) }
+
+    describe 'date_from' do
+      it { is_expected.to validate_presence_of(:date_from) }
+
+      context 'when the date_from is less than date_to' do
+        before { report.date_from = Time.zone.tomorrow }
+
+        it { expect(subject.valid?).not_to be true }
```
novotnyjakub: This could read slightly better, I think:

```ruby
it { is_expected.not_to be_valid }
```

app/controllers/outputs_controller.rb
```diff
@@ -0,0 +1,26 @@
+class OutputsController < ApplicationController
```
novotnyjakub: I'm not sure about this name, `OutputsController` seems quite ambiguous. What about `ReportsController`?

novotnyjakub:
:+1: looks good now, I'd merge. I've merged #509 now, so should we rebase on master and create new PR for this? Or maybe we can just merge in..

colinbruce:
I'll create a new PR on a rebased version.

PR #516, opened 2015-12-21
Finance report

colinbruce:
Allow admin users to output the finance data from between two dates to CSV

After login this will be accessible at outputs/finance_report

Updated following the merge of decision columns.

[Ticket](https://www.pivotaltracker.com/story/show/93418134)

[finishes #93418134]


novotnyjakub:
:+1:

PR #520, opened 2015-12-29
Fix for married applicant savings validation

colinbruce:
When a user processes a married applicants form and submitted
the savings and investment view without answering the first
question, a 500 error was raised.

This improves the logic to invalidate the object and render
the validation message on the view.

[finishes #110493440]


novotnyjakub:
:+1: great fix!

PR #521, opened 2015-12-29
Hide office BE Codes from managers

colinbruce:
When editing an office, managers should not see the office entity code
or be able to edit it.

Updated the tests for new office to match

[finishes #110499670]

[Ticket](https://www.pivotaltracker.com/story/show/110499670)


novotnyjakub:
:+1: good to go.

PR #524, opened 2015-12-31
Rake task to execute a local sql file

colinbruce:
This will allow users to scp a file to the sever and execute it without having to launch sql.
It replaces a ~/ with ENV['HOME'] so the file can be referenced without the absolute path.
The task is designed to abort gracefully if the file is missing, or not suffixed as .sql
It runs as a transaction so that, if an individual command fails, none will be committed. 

Usage:
rake run_sql:from_file[~/path/filename.sql]

### DIFF
--- a/lib/tasks/run_sql.rake
+++ b/lib/tasks/run_sql.rake
@@ -1,0 +1,22 @@
+namespace :run_sql do
+  desc 'Execute local file, save your file with a .sql extension'
+  task :from_file, [:file] => :environment do |_t, args|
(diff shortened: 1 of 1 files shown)


dotemacs:
Cool, but why?

colinbruce:
@dotemacs we have a lot of manual updates to make to the DB and we've been discussing options on how to automate/simplify the process.

This was a conversation starter, we can have a chat next week...

dotemacs:
Hello 

No problem. I just wanted to know the reasoning behind it.

Thanks

PR #527, opened 2016-01-04
Update column layout of confirmation page

colinbruce:
Minor fix for column layout


colinbruce:
[finishes #110564568]

dotemacs:
YEEEEES! I :heart: it so much!

PR #528, opened 2016-01-06
Show BEC on office edit view

dotemacs:
Added the BEC to show on the checkboxes when editing the office.

### DIFF
--- a/app/controllers/offices_controller.rb
+++ b/app/controllers/offices_controller.rb
@@ -22,6 +22,7 @@ class OfficesController < ApplicationController
   end
 
   def edit
+    @becs = Hash[@office.business_entities.map { |be| [be.jurisdiction_id, be.code] }]
   end
 
   def create
--- a/app/models/jurisdiction.rb
+++ b/app/models/jurisdiction.rb
@@ -3,8 +3,6 @@ class Jurisdiction < ActiveRecord::Base
   has_many :office_jurisdictions
   has_many :offices, through: :office_jurisdictions
 
-  has_many :business_entities
-
   validates :name, uniqueness: true, presence: true
   validates :abbr, uniqueness: { allow_nil: true }
 
--- a/app/models/jurisdiction.rb
+++ b/app/models/jurisdiction.rb
@@ -3,6 +3,8 @@ class Jurisdiction < ActiveRecord::Base
   has_many :office_jurisdictions
   has_many :offices, through: :office_jurisdictions
 
+  has_many :business_entities
+
   validates :name, uniqueness: true, presence: true
   validates :abbr, uniqueness: { allow_nil: true }
 
--- a/app/views/offices/_form.html.slim
+++ b/app/views/offices/_form.html.slim
@@ -27,6 +27,6 @@
             .columns.small-12.end
               = b.label do
                 = b.check_box
-                = b.text
+                = "#{b.text} (#{@becs[b.value]})"
 
   .actions = f.submit class: 'button'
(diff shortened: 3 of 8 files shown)


app/views/offices/_form.html.slim
```diff
@@ -27,6 +27,6 @@
             .columns.small-12.end
               = b.label do
                 = b.check_box
-                = b.text
+                = "#{b.text} (#{@becs.where(jurisdiction_id: b.value, office_id: @office.id).first.code})"
```
colinbruce: This shouldn't need the 
```
office_id: @office_id
```
anymore as it loads by office_id in the controller

app/models/jurisdiction.rb
```diff
@@ -3,6 +3,8 @@ class Jurisdiction < ActiveRecord::Base
   has_many :office_jurisdictions
   has_many :offices, through: :office_jurisdictions
 
+  has_many :business_entities
```
colinbruce: Do we *need* this here and on office?

The join is only used in tests, not in the code.

app/views/offices/_form.html.slim
```diff
@@ -27,6 +27,6 @@
             .columns.small-12.end
               = b.label do
                 = b.check_box
-                = b.text
+                = "#{b.text} (#{@becs.where(jurisdiction_id: b.value).first.code})"
```
novotnyjakub: I think we should avoid doing sql queries in the views at any cost. So ideally the `@becs` would either be pre-processed or the objects in the `@jurisdictions` array would somehow be decorated and include the `bec` straight away (not sure exactly how).

app/controllers/offices_controller.rb
```diff
@@ -22,6 +22,7 @@ def new
   end
 
   def edit
+    @becs = BusinessEntity.where(office_id: @office.id)
```
novotnyjakub: Here you could process the entities to give you a hash of `jurisdiction_id -> bec`, so that you can avoid the sql in the view (commented bellow):

```ruby
@becs = Hash[@office.business_entities.map { |be| [be.jurisdiction_id, be.code] ]
```

spec/features/manager_bec_details_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    login_as(manager)
+    visit office_path(manager.office)
+
+    expect(page).to have_text 'Office details'
+
+    click_link 'Change details'
+
+    manager.office.jurisdictions.each do |jurisdiction|
+      jurisdiction.business_entities.pluck(:code) do |code|
+        expect(page).to have_text code
```
novotnyjakub: maybe we should test the code is correctly displayed next to the jurisdiction? 

```ruby
expect(page).to have_text("#{jurisdiction.name} (#{code})")
```

app/views/offices/_form.html.slim
dotemacs: > I think we should avoid doing sql queries in the views at any cost.

Fully agree.

> (not sure exactly how)

The same here, hence the 'SQL in the views'.

If you can come up with a good clean way of doing it, I'd love to know.

The reason I left this filth in, is because I couldn't figure out a clean solution and figured that this is such a small query that it can be left in. But since I don't know of a cleaner way, I'd love to hear all the suggestions of doing it better.

spec/features/manager_bec_details_spec.rb
dotemacs: Good point. Will push it now!

app/controllers/offices_controller.rb
dotemacs: Great! Just what I needed. Pushing it now!

app/models/jurisdiction.rb
novotnyjakub: Yeah, I think we won't be needing this, because we don't need to look at all business entity codes for a jurisdiction.

novotnyjakub:
Code looks good! :+1: 
Tests failing atm though and Codeclimate doesn't seem to be happy (but I think it's just stuck or something so needs re-run).

PR #529, opened 2016-01-07
Improve authorisation

novotnyjakub:
Migration from `cancancan` to `Pundit` and most importantly making sure all controller actions are authenticated and authorised. 

Skipping authentication and authorisation is now an explicit action on case by case basis. 

In terms of the permissions for `Application`, at this point they're quite simplified. First of all we share the same `ApplicationPolicy` for the main application flow, but also for processed and deleted applications. Which we may need to split in the future, somehow. Also the main flow doesn't have a specific permission for each action (ie. `personal_details?`, `personal_details_save?`, ...) but instead of groups them until a single `update?`. Again, it's enough granularity for the moment, but maybe it'll be needed to allow access to the summary page to someone who doesn't have `update?` permission. We'll see about that. 

Permissions in general were quite tightened, so we may see that some users won't be able to do what they used to. That may be either that this PR made some wrong assumptions and made it too strict, or that they created a process using some wholes in the authorisation (ie. sending links between offices, ...).
For example `admin` now can't do anything with applications, create them or even see them. Users and managers can only strictly manipulate applications and related records for their office, etc.

There are some bits in the code that I intentionally left to be fixed later, because it's already too big as is. For example some objects are not authorised for `update?` after they have been changed, which is due to the fact that we're saving changes inside the form objects, which makes it much harder and needs to be refactored further. In some templates, we may beed to check for permissions before displaying certain links.


app/controllers/processed_applications_controller.rb
```diff
@@ -1,31 +1,29 @@
...
 
   def show
+    authorize application
+
     @form = Forms::Application::Delete.new(application)
     assign_views
   end
 
   def update
+    # TODO: The authorisation should be done after the attributes have been updated
```
dotemacs: Is this reminder for this PR or a later PR?

Gemfile
```diff
@@ -15,6 +15,7 @@ gem 'devise'
 gem 'devise_invitable'
 # roles
 gem 'cancancan', '~> 1.10'
+gem 'pundit', '~> 1.0'
```
colinbruce: Does pundit replace cancancan completely? Can we remove the unused gem?

colinbruce:
Looks good to me :+1:  :checkered_flag: Merge at will!

app/controllers/processed_applications_controller.rb
novotnyjakub: Yep, it is. It's related to the fact that we're updating the `Application` inside the form object and therefore we can't easily do authorisation after the attributes have been assigned, but before saved.

dotemacs:
Merge it when you feel ready.

colinbruce:
:tada:

PR #531, opened 2016-01-11
Remove income_result action

dotemacs:
Remove income_result controller action and view from the application
flow.

This is in the response to the court staff not completing the
application to the end and just 'walking off' once they get 'partial'
feedback from the app.

Addressing this ticket:
https://www.pivotaltracker.com/projects/1212124/stories/110976496


kelliematheson:
Tested on branch builder.
income= true
income = part-payment
income = false
are all generating the expected behaviour

Accepted, please merge.

colinbruce:
Shouldn't the view file be deleted too?

novotnyjakub:
Looks good, but can we please merge after the authorisation? It's already conflicting with `master` and I don't want to add more to it..

dotemacs:
@novotnyjakub ok, no problem, will merge after your PR

PR #533, opened 2016-01-11
Upgrade Ruby to 2.3.0

dotemacs:
Upgrading Ruby to 2.3.0 as it's the latest stable.

But really, I want to test the new Docker image building!
(Also updating it since the Docker image building is now using the same
version as the Ruby version instead of the super crazy MoJ specific Ruby
versioning)


colinbruce:
Looks good, merge at will.

PR #534, opened 2016-01-12
Update business entity

colinbruce:
This sets up the changes need for the discussed BE GUI.

It adds valid-from and valid-to fields, with validation for sequence, 
and creates a scope so that an office can request it's current 
business-entities.

### Merge 3 of 3 - into master


dotemacs:
Looks good, go!

app/models/business_entity.rb
```diff
@@ -2,5 +2,15 @@ class BusinessEntity < ActiveRecord::Base
...
   belongs_to :jurisdiction
 
-  validates :office, :jurisdiction, :code, :name, presence: true
+  validates :office, :jurisdiction, :code, :name, :valid_from, presence: true
+
+  validates :valid_to, date: {
+    after: :valid_from, allow_blank: true
+  }
+
+  scope :current_by_office, lambda { |office_id|
```
novotnyjakub: I'm thinking that this should either be a `default` scope, or for example the `Office#business_entities` should use this scope. I'm not sure what the right solution is.

db/migrate/20160112092557_add_valid_dates_to_business_entity.rb
```diff
@@ -0,0 +1,16 @@
+class AddValidDatesToBusinessEntity < ActiveRecord::Migration
+  def up
+    add_column :business_entities, :valid_from, :datetime
+    add_column :business_entities, :valid_to, :datetime
+    remove_index :business_entities, name: :unique_office_jurisdiction
+    add_index :business_entities, [:office_id, :jurisdiction_id, :valid_to], name: :unique_active_office_jurisdiction
```
novotnyjakub: This is interesting, why only the `valid_to` column?

db/migrate/20160112092557_add_valid_dates_to_business_entity.rb
```diff
@@ -0,0 +1,16 @@
...
+    execute('UPDATE business_entities SET valid_from = created_at')
```
novotnyjakub: I think we'll have to set the `valid_from` to an older date, won't we? Something like the oldest application creation date. Because if we want to generate the financial report, it will only work on date ranges of valid business entities and therefore it would ignore some. Or won't it?

app/services/reference_generator.rb
```diff
@@ -14,7 +14,7 @@ def attributes
 
   def business_entity
     @business_entity ||= BusinessEntity.where(
```
novotnyjakub: I think here we'd ideally have some method on `BusinessEntity` to give us a valid one for a particular `office_id` and `jurisdiction_id`, so that we don't have to repeat those queries outside `BusinessEntity`. The valid dates are internal thing to this object and are not concern of most parts of the app.

app/models/office.rb
```diff
@@ -15,4 +15,10 @@ class Office < ActiveRecord::Base
   def managers
     users.where(office_id: id, role: 'manager')
   end
+
+  def business_entities
```
novotnyjakub: This now gives us currently valid `BusinessEntity` objects, which is fine. Will it be a problem when we want to get all of them, including the old ones (ie. for the financial report)?

db/migrate/20160112092557_add_valid_dates_to_business_entity.rb
colinbruce: I added the minimum needed! [:office_id, :jurisdiction_id] was insufficient, adding valid_to ensured that an office/jurisdiction combination could be unique...

app/models/office.rb
colinbruce: The financial report shouldn't be an issue, as it loads applications and then reads back to the jurisdiction > business_entity.

app/services/reference_generator.rb
colinbruce: Noted... Will add to the list of things to look into.

db/migrate/20160112092557_add_valid_dates_to_business_entity.rb
colinbruce: Good point.  We could just use the first created date from applications.created_at!

novotnyjakub: Yep, something fixed in the migration is fine.

app/services/reference_generator.rb
novotnyjakub: :+1:

app/models/office.rb
novotnyjakub: Good stuff, we should probably just later extend our tests on the financial report side to include such a case and we're good!

db/migrate/20160112092557_add_valid_dates_to_business_entity.rb
novotnyjakub: Cool, sounds good.

GitCop:

 * Commit: 8a46bd2b720962d4e94ad15b8276acacb0616a83

---

novotnyjakub:
Looks good! :+1: Thanks for the changes.

GitCop:

 * Commit: 8b1d1327842f99a83e953aa4e90968929ffb788e

---

PR #535, opened 2016-01-12
use cookie to autofill username

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/100523714)

* When user submits the login form, set a cookie to the content of the username (email) field
* When loading the login form, check for that cookie and if present, prefill the username field
* Prevent browser autocomplete of password box using decoy visually hidden password field (`autocomplete="off"` is not respected by most browsers any more)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,6 +21,8 @@
 //= require applicants_partner_over_61
 //= require evidence_confirmation
 //= require guide
+//= require cookies
+//= require login
 
 $(function(){
   Foundation.global.namespace = '';
--- a/app/assets/javascripts/cookies.js
+++ b/app/assets/javascripts/cookies.js
@@ -1,0 +1,47 @@
+var FR = FR || {};
+
+FR.Cookies = {
+  set: function (name, value, options){
+    if (typeof options === 'undefined') {
+      options = {};
+    }
+    var cookieString = name + '=' + value + '; path=/',
+        date;
+    if (options.days) {
+      date = new Date();
+      date.setTime(date.getTime() + (options.days * 24 * 60 * 60 * 1000));
+      cookieString = cookieString + '; expires=' + date.toGMTString();
+    }
+    if (document.location.protocol === 'https:') {
+      cookieString = cookieString + '; Secure';
+    }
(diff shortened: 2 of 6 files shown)


dotemacs:
I can add a feature test for this tomorrow!

dotemacs:
Another thing: **fr** (Fee Remission) name use to be relevant before, the app is now referred to as **Help With Fees**.

axemonkey:
Better yet - you can educate me in how to add a feature test!

dotemacs:
@axemonkey cool, so a good starting point would be to look at this simple feature and borrow the relevant bits: 

https://github.com/ministryofjustice/fr-staffapp/blob/master/spec/features/user_can_record_feedback_spec.rb

I'm guessing a good test for the feature you've implemented is to
 
1. log in as a user
1. log out
1. go back to the login page and check if the user email is populated

Let me know if you'd like some helpful nudging

axemonkey:
@dotemacs Rock, sounds good.

axemonkey:
Still writing a feature test atmo, so don't merge...

novotnyjakub:
Have we confirmed that we can store user's e-mail in the unencrypted cookies in the first place? It is a personal information.

axemonkey:
Good point - it's got to be stored locally somehow for this ticket to work, since Firefox doesn't autofill. Perhaps we could add a "Remember me" checkbox, to make the user opt in?

novotnyjakub:
Yep and then we could store it in Rails session somehow, rather than a cookie. The Rails session is a cookie as well, but it's encrypted, so at least it can't be read. Even though theoretically it can be hijacked.

dotemacs:
One point to remember, this app is only accessible on DOM1 network. So if somebody tinkers with the cookie, they've already gained access to DOM1 and at that point we've got bigger security implications to worry about...

axemonkey:
@dotemacs Good point about DOM1. Unencrypted cookie should be fine then. I guess we'll need some confirmation that this will pass assessment though?

Re: feature testing - we have minimal JS, and some of it is untested. I'd like to use something like Nightwatch to do functional testing on JS functionality instead of trying to bend Capybara to our will.

jitenp:
Passed testing.

PR #536, opened 2016-01-13
**Abandoned** Allow business entity editing

colinbruce:
This adds methods for admins to see the current state of the business entities for
a given office, add a new business entity and update an existing business_entity.code and .name.

When updating *only* a BE.name, no new item is created.  If the code is amended, the existing BE has it's valid to date set and a new BE is created with the same start date and the new code.


app/views/business_entities/index.html.slim
```diff
@@ -0,0 +1,32 @@
...
+
+#result_table
+  .row
+    .columns.small-3.bold Jurisdiction
+    .columns.small-2.bold BEC
+    .columns.small-4.bold.end BEC Description
+  -@jurisdictions.each do |j|
+    .row
+      .columns.small-3
+        =j['jurisdictions.id']
```
colinbruce: This is an error and is fixed in the #551 PR, it should have been name.

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def valid_business_entity
+    be = BusinessEntity.find_by(office_id: params[:office_id],
+                                jurisdiction_id: params[:jurisdiction_id],
+                                valid_to: nil)
+    flash[:alert] = t('error_messages.create_be_exists', jurisdiction: be.jurisdiction.name,
+                                                         office: be.office.name) if be
+    be
+  end
```
dotemacs: Here, I'd use **business_entity** instead of **be**. Partly because this is 'hidden' like here:

```ruby
flash[:alert] = t('error_messages.create_be_exists', jurisdiction: be.jurisdiction.name,
                                                     office: be.office.name) if be
```

Just because this line is so long, I'd maybe write it as:

```ruby
flash[:alert] = t('error_messages.create_be_exists', 
                         jurisdiction: be.jurisdiction.name,
                         office: be.office.name) if business_entity
```
But looking at it now, it also seems to 'hide' the **if** condition.

This might seem better:

```ruby
if business_entity
  flash[:alert] = t('...
end
business_entity
```

What do you think?

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    authorize business_entity
+  end
+
+  def update
+    bes = business_entity_service
+    new_be = bes.check_update(business_entity_params)
+    authorize new_be
+
+    if bes.persist_update!(new_be)
+      redirect_to office_business_entities_path
```
dotemacs: Personally, I'd avoid the abbreviations of the variables and I'd call them by what they are.

Also, do we need to have **bes** variable at all? Could this:

```ruby
def update
  bes = business_entity_service
  new_be = bes.check_update(business_entity_params)
  authorize new_be

  if bes.persist_update!(new_be)
```
be written as:

```ruby
def update
  new_business_entity = business_entity_service.check_update(business_entity_params)
  authorize new_business_entity

  if business_entity_service.persist_update!(new_business_entity)
```

The reason I'm asking to avoid the abbreviated variable names is because I think the second example 'reads' easier. There is no need for the 'translation' as to what the abbreviations stand for.

What do you think?

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    flash[:alert] = t('error_messages.create_be_exists', jurisdiction: be.jurisdiction.name,
+                                                         office: be.office.name) if be
+    be
+  end
+
+  helper_method def jurisdiction
+    j_id = find_jurisdiction_id
+    return false unless j_id
+    @jurisdiction ||= Jurisdiction.find(j_id)
+  end
```
dotemacs: Again, I'd avoid the use of the abbreviated variable.

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def business_entity_service
+    jurisdiction
+    BusinessEntityService.new(office, @jurisdiction)
+  end
+
+  def list_jurisdictions
+    @jurisdictions = Jurisdiction.joins(join(office.id)).order(order_sequence).
+                     pluck_h(list_fields).each { |j| j[:state] = state(j) }
```
dotemacs: This seems confusing to me. It doesn't read easily.

If possible I'd move it to a 'Query Object' and store it under **app/models/query** that would do this one thing well.

The use of **pluck_h** is also a bit strange and leads to head scratching.

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+  def join(office_id)
+    <<-JOIN
+      LEFT OUTER JOIN business_entities ON business_entities.jurisdiction_id = jurisdictions.id
+        AND business_entities.office_id = #{office_id} AND business_entities.valid_to IS NULL
+      LEFT OUTER JOIN office_jurisdictions
+        ON business_entities.jurisdiction_id = office_jurisdictions.jurisdiction_id
+          AND business_entities.office_id = office_jurisdictions.office_id
+          AND business_entities.valid_to IS NULL
+    JOIN
+  end
```
dotemacs: This method begs to be a 'Query Object' and extracted, especially because it's in a controller.

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    if j['office_jurisdictions.office_id'].present?
+      'edit'
+    else
+      j['business_entities.code'].present? ? 'delete' : 'new'
+    end
+  end
+
+  def order_sequence
+    'business_entities.code IS NOT NULL DESC, jurisdictions.id'
+  end
```
dotemacs: Maybe this can be a scope in a model?

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    jurisdiction
+    BusinessEntityService.new(office, @jurisdiction)
+  end
+
+  def list_jurisdictions
+    @jurisdictions = Jurisdiction.joins(join(office.id)).order(order_sequence).
+                     pluck_h(list_fields).each { |j| j[:state] = state(j) }
+  end
+
+  def state(j)
```
dotemacs: I find the use of the single letter as a parameter confusing.

app/models/jurisdiction.rb
```diff
@@ -20,4 +20,8 @@ def display_full
     result.concat(" (#{abbr})") unless abbr.blank?
     result
   end
+
+  def self.pluck_h(keys)
```
dotemacs: Why the name **pluck_h**?
Maybe it should have a name that describe what is meant to do?

app/models/office.rb
```diff
@@ -13,4 +13,10 @@ class Office < ActiveRecord::Base
   def managers
     users.where(office_id: id, role: 'manager')
   end
+
+  def business_entities
+    BusinessEntity.where(office_id: id).
+      where('valid_to IS NULL').
+      order(:jurisdiction_id)
```
dotemacs: Maybe this can be a query object also?

We've made similar queries as query objects.

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def check_update(params)
+    build_business_entity(params)
+  end
+
+  def persist_update!(business_entity)
+    return false if business_entity.nil? || @business_entity.nil?
+    if duplicate_needed?(business_entity)
+      @business_entity.assign_attributes(valid_to: @timestamp)
+      save_entities_in_transaction(business_entity)
```
dotemacs: I'm not familiar with the requirement, hence the question: why is there a need to save duplicate:
* @business_entity
* business_entity

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,56 @@
...
+  def duplicate_needed?(business_entity)
+    business_entity != @business_entity &&
+      (both_entities_present?(business_entity) && business_entity.code != @business_entity.code)
+  end
+
+  def both_entities_present?(business_entity)
+    @business_entity.present? && business_entity.present?
+  end
+
+  def save_entities_in_transaction(business_entity)
```
dotemacs: Maybe a name of **save_duplicate** instead of **save_entities_in_transaction**?

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,56 @@
...
+                       jurisdiction: @jurisdiction,
+                       name: params[:name],
+                       code: params[:code],
+                       valid_from: @timestamp)
+  end
+
+  def duplicate_needed?(business_entity)
+    business_entity != @business_entity &&
+      (both_entities_present?(business_entity) && business_entity.code != @business_entity.code)
+  end
```
dotemacs: This is a bit confusing as it doesn't read easily. I'd break it down into smaller chunks.

Maybe something like this:

```ruby
def duplicate_needed?(business_entity)
  entities_match?(business_entity) &&
    both_entities_present_and_codes_match?(business_entity)
end

def entities_match?(business_entity)
  business_entity != @business_entity
end

def codes_match?(business_entity)
  business_entity.code != @business_entity.code
end

def both_entities_present_and_codes_match?(business_entity)
  both_entities_present?(business_entity) && codes_match?(business_entity)
end
```

app/controllers/business_entities_controller.rb
novotnyjakub: Yep, I think `business_entity_service` could be a private method returning `@business_entity_service ||= ...`.

I see why @colinbruce is using abbreviations, I use them sometimes because of the line length restriction, which I find annoying at times.

app/controllers/business_entities_controller.rb
novotnyjakub: As per the comment on #534 Ideally we'd get a method on `BusinessEntity`, which would return a valid one for given `office` and `jurisdiction`. Then this could be as simple as:

```ruby
def valid_business_entity
  BusinessEntity.valid_for(office_id: params[:office_id], jurisdiction_id: params[:jurisdiction_id]).tap do |be|
    flash[:alert] = t('error_messages.create_be_exists', jurisdiction: be.jurisdiction.name, office: be.office.name) if be
  end
end
```

I personally don't have a problem with an abbreviated block parameter name, but I'll leave it up to you. 

Maybe the flash alert could be abstracted to a private method, so then it would be 

```ruby
flash[:alert] = business_entity_exists_alert(be) if be
```

app/controllers/business_entities_controller.rb
novotnyjakub: I'd move it to the `Jurisdiction` model as a scope/class method or a query model. Just make it return what's expected in this controller. Then all the supporting methods bellow should be there too and the controller doesn't generate any SQL?

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,56 @@
+class BusinessEntityService
```
novotnyjakub: I think it's quite solid, if you look at @dotemacs 's suggestions bellow, it'll be more readable, I agree. 

The only extra suggestion I have is to change the name of `@business_entity` to something like `@existing_business_entity`. So that it's clear especially when we're saving both in transaction.

app/views/offices/edit.html.slim
```diff
@@ -3,6 +3,10 @@ header
 
 = render 'form', with_jurisdictions: true
 
+- if policy(:business_entity)
```
novotnyjakub: Didn't know that this would work, does this infer `.index?`?

novotnyjakub:
This is a huge amount of code, sorry about all the comments. One thing I find missing though is some feature tests, to cover adding / editing of BECs.

app/controllers/business_entities_controller.rb
dotemacs: > Ideally we'd get a method on BusinessEntity...

I'd argue that ideally we'd put it in a Query Object, as that is a path we're using already.

app/controllers/business_entities_controller.rb
dotemacs: > I see why @colinbruce is using abbreviations, I use them sometimes because of the line length restriction, which I find annoying at times.

True, they are annoying. But I feel that those are the constraints placed in order to enforce better discipline when developing. There are always ways to work around the line length, without sacrificing readability.

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    @jurisdiction ||= Jurisdiction.find(j_id)
+  end
+
+  def find_jurisdiction_id
+    return params[:jurisdiction_id] if params[:jurisdiction_id].present?
+    return business_entity.jurisdiction_id if params[:id].present?
+  end
+
+  def business_entity_service
+    jurisdiction
```
dotemacs: I'm a bit confused as to why this is called on its own?

app/services/business_entity_service.rb
dotemacs: Good point by @novotnyjakub. I was really worried that I wasn't getting the existence of both:

* @business_entity 
* business_entity

Renaming them into something so that they differ would be much easier to read and follow.

app/models/jurisdiction.rb
colinbruce: Because instead converting **to a hash** with **to_h** I want to **pluck a hash** with pluck_h.

Plus it was the example I found on-line! ;)

But I am going to look at re-writing this whole block anyway...

app/services/business_entity_service.rb
colinbruce: Because we are creating a new business_entity while the existing one is *deactivated*. If saving one or other fails, the transaction protects the DB integrity

colinbruce:
Due to the scope and nature of the changes, along with the need to merge to the source BE branch, this PR will be closed.  A new branch has been created a new PR will be created for separate review.  

I'll leave this here until the new version is merged, as a reference.

PR #537, opened 2016-01-13
remove turbolinks, because it is horrid

axemonkey:


novotnyjakub:
I like the title :-D But what are the implications of this, please? Are we using it at all?

axemonkey:
Turbolinks, as I'm sure you know, turns all regular page href links into (theoretically faster - this is questionable) AJAX links and replaces the entirety of `<body>` rather than loading the whole page again. This messes with JS functions that run on load (you can get round this by watching the DOM for changes, but it's a crappy workaround for dubious benefit.)

Additionally, I noticed after removing `jquery.turbolinks` from `application.js` that `turbolinks` itself was absent, so no, we weren't even using it. This is effectively just removing extra cruft.

novotnyjakub:
Cool, thanks for this. I'm happy with this :+1:

PR #539, opened 2016-01-14
Import bec from spreadsheet

novotnyjakub:
This adds 2 rake tasks which accept a `csv` as an input. The tasks are:

* `bec_import:prune`, which deletes unused `BusinessEntity` objects from the database. Unused means they 1) are not in the given csv, 2) no office has jurisdictions assigned with it, 3) no application has ever been processed with it.
* `bec_import:update`, which updates existing `BusinessEntity` objects based on the csv. It updates the `code` if available and `name` if available.

There are some outstanding UI changes to be done before the import should actually run in production.


lib/bec_import.rb
```diff
@@ -0,0 +1,57 @@
+class BecImport
+  def initialize(lines)
+    @lines = lines
+    validate!
+  end
+
+  def delete_unused
+    BusinessEntity.all.each do |be|
+      be.destroy if to_delete?(be)
```
colinbruce: What is going to happen if an office has changed their jurisdictions since the spreadsheets were generated?

colinbruce:
Looks fine! :+1:

PR #540, opened 2016-01-15
Allow users to edit their own passwords

dotemacs:
Since the introduction of Pundit, the users were unable to change their
own passwords. This change allows them to do that again.

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -1,5 +1,12 @@
 module Users
   class RegistrationsController < Devise::RegistrationsController
+
+    before_action :authorize_for_current_user
+
+    def edit
+      super
+    end
+
     def update
       if account_update_params[:password].present?
         super
@@ -8,6 +15,16 @@ module Users
         resource.errors.add(:password, error_message)
         respond_with resource
       end
+    end
+
+    private
+
+    def their_own?
+      current_user.id.eql?(params[:id].to_i)
+    end
+
+    def authorize_for_current_user
+      authorize current_user if their_own?
     end
   end
 end
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -1,13 +1,13 @@
 module Users
   class RegistrationsController < Devise::RegistrationsController
 
-    before_action :authorize_for_current_user
-
     def edit
+      authorize User.find(params[:id]), :edit_password?
       super
     end
 
     def update
+      authorize User.find(params[:id]), :update_password?
       if account_update_params[:password].present?
         super
       else
@@ -15,16 +15,6 @@ module Users
         resource.errors.add(:password, error_message)
         respond_with resource
       end
-    end
-
-    private
-
-    def their_own?
-      current_user.id.eql?(params[:id].to_i)
-    end
-
-    def authorize_for_current_user
-      authorize current_user if their_own?
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -1,5 +1,11 @@
 module Users
   class RegistrationsController < Devise::RegistrationsController
+
+    def edit
+      authorize current_user if their_own?
```
dotemacs: I'd like some suggestions on how handle this better.

app/controllers/users/registrations_controller.rb
```diff
@@ -9,5 +15,11 @@ def update
         respond_with resource
       end
     end
+
+    private
+
+    def their_own?
```
dotemacs: is this name descriptive enough?

colinbruce:
Looks okay to me, might benefit from @novotnyjakub checking the pundit usage!

app/controllers/users/registrations_controller.rb
novotnyjakub: This is what the `UserPolicy` responsibility is, so it should only be something like this:

```ruby
authorize current_user, :edit_password?
```

And then in the `UserPolicy`:

```ruby
class UserPolicy < BasePolicy
  ....
  def edit_password?
    user_themselves?
  end
  ....
end
```

Plus specs for the policy.

novotnyjakub: Or maybe it can be just called `edit?` instead of `edit_password?`, I'm not sure if the edit function has also other responsibilities.

app/controllers/users/registrations_controller.rb
```diff
@@ -9,5 +16,15 @@ def update
...
+      current_user.id.eql?(params[:id].to_i)
+    end
+
+    def authorize_for_current_user
+      authorize current_user if their_own?
```
novotnyjakub: Actually looking at this again, authorising `current_user` doesn't make sense, cause this will always be true. We need somehow run `authorize user` and the `user` has to be the one being changed. 

So something like this:

```ruby
user = User.find(params[:id])
authorize user, :edit_password?
```

app/controllers/users/registrations_controller.rb
novotnyjakub: This is already in the `UserPolicy` as per the comment bellow.

app/controllers/users/registrations_controller.rb
dotemacs: But by doing that, this test breaks:

https://github.com/ministryofjustice/fr-staffapp/blob/user-edit-own-password/spec/features/user_profile_spec.rb#L44

dotemacs: The idea is to allow currently logged in users to edit their own passwords only.

novotnyjakub: Yep, but it's the policy who has to decide that. So the `UserPolicy` should have something like this:

```
def edit_password?
  user_themselves?
end
```

And then you're authorising a `user` (which comes from the params) in context of the `current_user`, which then only returns `true` when it's the same user.

app/controllers/users/registrations_controller.rb
```diff
@@ -1,5 +1,11 @@
 module Users
   class RegistrationsController < Devise::RegistrationsController
+
+    def edit
+      authorize User.find(params[:id]), :edit_password?
```
novotnyjakub: Cool, this looks good. We should do the same in the `update` action. Arguably we should have `update_password?` policy method for that one, but we can get away with reusing this one.

app/policies/user_policy.rb
```diff
@@ -23,6 +23,10 @@ def create?
     (manager? && same_office? && !setting_to_admin_role?) || admin?
   end
 
+  def edit_password?
```
novotnyjakub: Missing test for this.

PR #542, opened 2016-01-15
Changes to Add new office page

novotnyjakub:
When admin tries to add a new office, now it only requires a name. 

This PR also removes the validation on `Office` to have at least one `OfficeJurisdiction`. 
To prevent staff creating applications and then not being able to progress through application details, the `Start now` button only appears when there is at least 1 jurisdiction. Otherwise there's a message to contact a manager.


dotemacs:
Looks OK

colinbruce:
Looks fine, :+1:

PR #544, opened 2016-01-15
process application module changes

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/110953430)

Before
---
screen shot 2016-01-15 at 16 23 15

After
---
screen shot 2016-01-15 at 16 23 30

Note the Pivotal story says to remove the space between the bulleted list and the 'Start now' button, but this looks too cramped. Removing the top margin from the button container has reduced the space, which looks good.

Here is the cramped version:
screen shot 2016-01-15 at 15 49 07

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -27,4 +27,5 @@
  *= require details
  *= require phase_banner
  *= require headings
+ *= require utilities
  */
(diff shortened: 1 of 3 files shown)


kelliematheson:
Agreed, looks perfect. Please merge when possible.

PR #546, opened 2016-01-18
Removing the Ruby version from Travis CI config

dotemacs:
Previously there was a reason for specifiying the Ruby version in both
.travis.yml and .ruby-version. Hoping that this is no longer the case,
since maintaining the Ruby version in two places is error prone.


dotemacs:
TravisCI builds based on .ruby-version

PR #548, opened 2016-01-18
hide staff guides in footer if logged out

axemonkey:
Teeny little PR so that you don't see links to the staff guides if you're logged out (since you can't view them when logged out).


colinbruce:
Approved for UX, pending review of logged out users being able to access the guide 

@leighmoney Can we chat next time you are in the office?

PR #549, opened 2016-01-18
Manager only sees jurisdictions with bec

novotnyjakub:
I'm creating this PR to be merged to #542 so that both can be released together, because they depend on each other in terms of UX. 

The big part of this PR is refactoring of the `office_controller_spec.rb` which is now made in a way that not all functionality has to be tested for all types of users. It's now driven by the policy, rather than being sort of an integration test, which is hard to maintain.

### DIFF
--- a/app/controllers/offices_controller.rb
+++ b/app/controllers/offices_controller.rb
@@ -42,8 +42,10 @@ class OfficesController < ApplicationController
     office.assign_attributes(office_params)
     authorize office
 
-    if office.save && manager_setup.in_progress?
-      redirect_to out_of_the_box_redirect
+    if office.save
+      flash[:notice] = 'Office was successfully updated'
+
+      redirect_to update_redirect_path
     else
       respond_with(office)
     end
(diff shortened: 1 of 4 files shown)


app/controllers/offices_controller.rb
```diff
@@ -42,8 +42,10 @@ def update
     office.assign_attributes(office_params)
     authorize office
 
-    if office.save && manager_setup.in_progress?
-      redirect_to out_of_the_box_redirect
+    if office.save
```
dotemacs: Just curious as to what happens when a manager signs in for a first time?

Before they'd be prompted to assign certain offices, but what happens now?

dotemacs:
From the code perspective, I don't see anything that jumps at me.

app/controllers/offices_controller.rb
colinbruce: It's now being handled in the update_redirect_path here: https://github.com/ministryofjustice/fr-staffapp/pull/549/files#diff-5ab24263464befc902ee5117ae8011beR72

colinbruce:
Looks fine to me :+1: Merge into the other office branch at will

PR #554, opened 2016-01-20
Enable BEC addition and updates

colinbruce:
# Replaces #536 

This adds methods for admins to see the current state of the business entities for
a given office, add a new business entity and update an existing business_entity.code and .name.

When updating only a BE.name, no new item is created. If the code is amended, the existing BE has it's valid to date set and a new BE is created with the same start date and the new code.

---

This PR address points raised in the comments on #536 and, as it is to be merged into the #534, eliminates the duplication of those files.

### Merge 2 of 3


GitCop:

 * Commit: 33544a47d36774f842bef101dce736a8152e6c76

---

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,65 @@
...
+  def persist_update!(business_entity)
+    return false if business_entity.nil? || @existing_business_entity.nil?
+    if duplicate_needed?(business_entity)
+      @existing_business_entity.assign_attributes(valid_to: @timestamp)
+      save_entities_in_transaction(business_entity)
+    else
+      @existing_business_entity.assign_attributes(name: business_entity.name)
+      @existing_business_entity.save
+    end
+  end
```
dotemacs: I'd consider doing this for this method:

```ruby  
def persist_update!(business_entity)
  return false if both_entities_nil?(business_entity)
  determine_duplication(business_entity)
end

# add this under private

def both_entities_nil?(business_entity)
  business_entity.nil? || @existing_business_entity.nil?
end

def determine_duplication(business_entity)
  if duplicate_needed?(business_entity)
    save_both_entities(business_entity)
  else
    update_existing_entity(business_entity)
  end
end

...

# rename save_entities_in_transaction
def save_both_entities(business_entity)
  @existing_business_entity.assign_attributes(valid_to: @timestamp)
  ActiveRecord::Base.transaction do
    @existing_business_entity.save
    business_entity.save
  end
end

def update_existing_entity(business_entity)
  @existing_business_entity.assign_attributes(name: business_entity.name)
  @existing_business_entity.save  
end
```

app/controllers/business_entities_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      flash[:alert] = t('error_messages.create_be_exists',
+        jurisdiction: business_entity.jurisdiction.name,
+        office: business_entity.office.name) if business_entity
+    end
+  end
+
+  helper_method def jurisdiction
+    jurisdiction_id = find_jurisdiction_id
+    return false unless jurisdiction_id
+    @jurisdiction ||= Jurisdiction.find(jurisdiction_id)
```
novotnyjakub: The controller looks neat now. A small change I'd suggest is to this method:

```ruby
Jurisdiction.find(jurisdiction_id) if jurisdiction_id
```

app/models/query/business_entity_management.rb
```diff
@@ -0,0 +1,13 @@
+module Query
+  class BusinessEntityManagement
+    def initialize(office)
+      @office = office
+    end
+
+    def jurisdictions
+      Jurisdiction.all.map do |jurisdiction|
+        Views::OfficeBusinessEntityState.new(@office, jurisdiction)
```
novotnyjakub: I'm not sure the query object should return view objects, but rather a `ActiveRecord::Relation`? Then the views creation is a responsibility of the controller.

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,72 @@
+class BusinessEntityService
```
novotnyjakub: This looks really good now!

app/models/views/office_business_entity_state.rb
```diff
@@ -0,0 +1,54 @@
+module Views
+  class OfficeBusinessEntityState
```
novotnyjakub: This is a really good abstraction. It's missing a spec, though?

colinbruce: @novotnyjakub do you mean that a specific feature is missing from [here](https://github.com/ministryofjustice/fr-staffapp/pull/554/files#diff-0de8e4919a5e91498a48885f86d0454bR4)

novotnyjakub: No, I meant that this class doesn't have a test, but I was just completely blind. My apologies!

app/models/query/business_entity_management.rb
colinbruce: So, for now the query object would be superfluous?  Because it would only return Jurisdiction.all and then the controller would map the View objects from it?

If you are in tomorrow, maybe it would be better to pair and discuss?

app/models/views/office_business_entity_state.rb
dotemacs: :see_no_evil:

app/services/business_entity_service.rb
```diff
@@ -0,0 +1,72 @@
...
+  def build_update(params)
+    @new_params = params
+    build_new_business_entity
+    @persist_status = set_update_type
+    if @persist_status.eql?(:update_existing)
+      @existing_business_entity.assign_attributes(name: @new_business_entity.name)
+      @existing_business_entity
+    else
+      @existing_business_entity.assign_attributes(valid_to: @timestamp)
+      @new_business_entity
```
dotemacs: Just to be pedantic, what if this whole `if...else` was done as:

```ruby
@persist_status.eql?(:update_existing) ? update_existing : return_new

private

def update_existing
  @existing_business_entity.assign_attributes(name: @new_business_entity.name)
  @existing_business_entity
end

def return_new
  @existing_business_entity.assign_attributes(valid_to: @timestamp)
  @new_business_entity  
end
```

colinbruce: Cool... I'll raise you a:-

```ruby
@persist_status.eql?(:update_existing) ? update_and_return_existing : update_existing_return_new

private

def update_and_return_existing
  @existing_business_entity.assign_attributes(name: @new_business_entity.name)
  @existing_business_entity
end

def update_existing_return_new
  @existing_business_entity.assign_attributes(valid_to: @timestamp)
  @new_business_entity  
end
```

app/models/query/business_entity_management.rb
novotnyjakub: Yep, you're right, there's no need for this query object. Originally the sql query was much more complicated, wasn't it. But now it's lean, so you can just make the controller construct this array.

PR #560, opened 2016-01-22
make green buttons consistent

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/112148431)

Seems to be caused by actually applying the class 'success' to some buttons, so that's been removed.

🚫
button-before

👌
button-after


kelliematheson:
Lovely, please merge

PR #563, opened 2016-01-25
Move graphs from admin/index

colinbruce:
As the number of offices has grown, the index page for admins has slowed
down more and more.

This PR aims to make it faster while still allowing relatively quick access to the data

I have migrated the graphs to the reports controller and 
provided a link to new action on front page

screen shot 2016-01-25 at 11 31 06


dotemacs:
It looks fine to me.

PR #565, opened 2016-01-25
Remove benefits result page

novotnyjakub:
To avoid incomplete applications, the result page has now been removed. When the applicant says they're on benefits and the result is overridable, the benefit override page is displayed straight away. 

This involved many changes, from the `BenefitOverridesController` to the UI and style. Hopefully nothing's missing.

It still needs a few visual treatments.


app/controllers/applications/process_controller.rb
```diff
@@ -148,13 +134,31 @@ def form_params(type)
...
+
+        if benefit_check_runner.can_override?
+          redirect_to application_benefit_override_paper_evidence_path(application)
+        else
+          redirect_to(action: :summary)
+        end
+      else
+        redirect_to(action: :income)
+      end
+    end
```
dotemacs: Just because this is a nested `if` statement I'd look into splitting it up:

```ruby
def benefit_check_and_redirect(benefits)
  if benefits
    benefit_check_runner.run
    determine_override
  else
    redirect_to(action: :income)
  end
end

def determine_override
  if benefit_check_runner.can_override?
    redirect_to application_benefit_override_paper_evidence_path(application)
  else
    redirect_to(action: :summary)
  end  
end
```

app/controllers/applications/process_controller.rb
```diff
@@ -148,13 +134,31 @@ def form_params(type)
...
+
     def calculate_income
```
dotemacs: I know that you haven't touched this method. But it could be 'shaved off' by moving the "whole content of the method" back to where `calculate_income` is called.

dotemacs:
There's two `privates`:

* https://github.com/ministryofjustice/fr-staffapp/blob/remove-benefits-result-page/app/controllers/applications/process_controller.rb#L123
* https://github.com/ministryofjustice/fr-staffapp/blob/remove-benefits-result-page/app/controllers/applications/process_controller.rb#L129

dotemacs:
Possibly out of scope of this PR, but this could maybe be made into it's own class, that just returns those instances:

https://github.com/ministryofjustice/fr-staffapp/blob/remove-benefits-result-page/app/controllers/applications/process_controller.rb#L102

app/controllers/benefit_overrides_controller.rb
```diff
@@ -29,10 +29,19 @@ def application
...
+    params.require(:benefit_override).permit(*Forms::BenefitsEvidence.permitted_attributes.keys)
+  end
+
+  helper_method def error_message_partial
+    @error_message_partial ||= case application.last_benefit_check.dwp_result.try(:downcase)
+                               when nil, 'undetermined'
+                                 'missing_details'
+                               when 'server unavailable', 'unspecified error'
+                                 'technical_error'
+                               end
```
dotemacs: This could potentially be written as:

```ruby
helper_method def error_message_partial
  result = application.last_benefit_check.dwp_result.try(:downcase)
  @error_message_partial ||= \
  { 'undetermined': 'missing_details',
    'server unavailable': 'technical_error',
    'unspecified error': 'technical_error'
  }[result] || 'missing_details'
end
```

novotnyjakub: I personally find the `case` version more readable and it's more efficient. @colinbruce which one do you prefer?

colinbruce: Personally I find the `case` more legible too.

novotnyjakub: thanks, I'm leaving that one as is then

novotnyjakub:
@dotemacs I've addressed your comments, apart from the one of splitting out to a different class. The reason is that I think we should potentially slice this controller to RESTful one with `edit` / `update` actions of some sort for each of the pages. Like the `BenefitOverrideController` is now part of the main journey, but it's a clean small controller with 2 actions.

Is that ok?

dotemacs:
@novotnyjakub sounds good, go ahead with this PR

PR #567, opened 2016-01-26
Introduce mi users

novotnyjakub:
This adds a new user role to the system. Users with this role should only be able to manage their own profile and generate finance reports.


dotemacs:
Looks fine to me.

PR #568, opened 2016-01-26
Deactivate Business Entity codes

colinbruce:
This is the technical implementation of the delete/deactivate feature.
Content updates may need a separate PR when terminology is agreed.

### Replaces #561

### DIFF
--- a/app/controllers/business_entities_controller.rb
+++ b/app/controllers/business_entities_controller.rb
@@ -26,13 +26,13 @@ class BusinessEntitiesController < ApplicationController
     persist_and_redirect(:edit)
   end
 
-  def confirm
+  def deactivate
     authorize business_entity
   end
 
-  def deactivate
+  def confirm_deactivate
     authorize business_entity_service.build_deactivate
-    persist_and_redirect(:confirm)
+    persist_and_redirect(:confirm_deactivate)
   end
 
   private
--- a/app/controllers/business_entities_controller.rb
+++ b/app/controllers/business_entities_controller.rb
@@ -14,11 +14,7 @@ class BusinessEntitiesController < ApplicationController
 
   def create
     authorize business_entity_service.build_new(business_entity_params)
-    if business_entity_service.persist!
-      redirect_to office_business_entities_path
-    else
-      render :new
-    end
+    persist_and_redirect(:new)
   end
 
   def edit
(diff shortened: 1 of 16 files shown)


dotemacs:
Looks fine

config/routes.rb
```diff
@@ -85,6 +85,8 @@
 
   resources :offices do
     resources :business_entities
+    get '/business_entities/:id/confirm' => 'business_entities#confirm', as: 'confirm_deactivate_business_entity'
+    post '/business_entities/:id/deactivate' => 'business_entities#deactivate', as: 'deactivate_business_entity'
```
novotnyjakub: These 2 rules are related to the resource and ideally would be nested in it: 

```ruby
resources :business_entities do
  get :confirm, on: :member
  post :deactivate, on: :member
end
```

Something between these lines.

app/controllers/business_entities_controller.rb
```diff
@@ -27,16 +23,28 @@ def edit
 
   def update
     authorize business_entity_service.build_update(business_entity_params)
+    persist_and_redirect(:edit)
+  end
+
+  def confirm
```
novotnyjakub: At first sight, this confused me as to what is it confirming? I'm not insisting, but wouldn't it make sense to be `confirm_deactivate` to be clear what the action does?

spec/controllers/business_entities_controller_spec.rb
```diff
@@ -130,4 +130,35 @@
...
+      sign_in admin
+      get :confirm, office_id: office.id, id: business_entity.id
+    end
+
+    it { is_expected.to have_http_status(:success) }
+
+    it { is_expected.to render_template(:confirm) }
+
+    it 'assigns the @business_entity variable' do
+      expect(assigns(:business_entity)).to be_a_kind_of BusinessEntity
```
novotnyjakub: Here I think you can test very specifically that `expect(assigns(:business_entity)).to eql(business_entity)`.

spec/services/business_entity_service_spec.rb
```diff
@@ -102,6 +102,24 @@
...
+
+    it 'returns a persisted object' do
+      expect(subject.persisted?).to be true
+    end
+
+    it 'returns a valid object' do
+      expect(subject.valid?).to be true
+    end
+
+    it 'has no ID' do
```
novotnyjakub: The test title is different from the test itself.

novotnyjakub:
A few small things, but overall it's nicely build on the previous PRs :+1:

PR #570, opened 2016-01-26
Introduce Nightwatch for JS functionality testing

axemonkey:
Introduces [Nightwatch](http://nightwatchjs.org/) for testing JS functionality (which is always a nightmare in Capybara etc, and can test things that aren't unit-testable i.e. intended browser functionality when using JS).

Notes (to be added to README if and when this is merged to master)

Installation
---
Make sure you have `nodejs >= 0.12` installed, then just `npm install` from the root directory.

Running the tests locally
---
Make sure the site is running on `localhost:3000` then in another tab just do `make test` or to watch the tests run automated in Firefox, do `make test-firefox`

To run a single spec (there are only two at the moment) add `spec=<specname>` to the command, e.g. 
```
make test spec=login
```


colinbruce:
Looks good, automated would be a plus! :)

axemonkey:
Yes, we'll need to do some setup on Jenkins/CI to spin up the Rails app then run `npm install && make test` or whatever. It's not part of the PR though. :+1:

colinbruce:
I've created an issue (#582) to remind us, so merge at will! :)

PR #574, opened 2016-01-28
BECs labelling update

colinbruce:
The `Delete` button took users to a `deactivate` view.

The index page columns was titled BEC description, the edit page called the field Name

Index column sequence was code > name, the edit view was name > code.

These have all been resolved.


novotnyjakub:
:+1: good to go!

PR #577, opened 2016-01-29
Restore graph data for admin

colinbruce:
During a rather too agressive refactoring, all of the code for
the collation of graph data was moved to the reports controller
and subsequently deleted when it was not needed.

This returns the two front page graphs to their correct state.

[finishes #112616861]

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -3,6 +3,7 @@ class HomeController < ApplicationController
 
   def index
     manager_setup_progress
+    load_graphs_for_admin
     load_waiting_applications
   end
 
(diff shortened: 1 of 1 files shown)


novotnyjakub:
Maybe we should rather try to tie this to the policies - ie. who can access graphs, rather than just admin, but we can change that later if needed. So go ahead!

colinbruce:
There is a policy(:report).graphs? that controls the display, I just stopped sending them data! :)

novotnyjakub:
Yes, but if we allowed the `mi` users to see graphs, they wouldn't because the data wouldn't be sent. That's what I meant. But we can extend it if/when that happens.

colinbruce:
:+1:

PR #578, opened 2016-01-29
Updated the office permissions

colinbruce:
Updated the office.index? policy by enabling it only for admin and mi users.

[finishes #112617015]

### DIFF
--- a/app/policies/office_policy.rb
+++ b/app/policies/office_policy.rb
@@ -4,7 +4,7 @@ class OfficePolicy < BasePolicy
   end
 
   def show?
-    not_staff? || (staff? && same_office?)
+    admin_or_mi? || (staff_or_manager? && same_office?)
   end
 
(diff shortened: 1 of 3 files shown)


novotnyjakub:
I think also the manager should not be able to see details of other than their own office, if I remember the details.

colinbruce:
That was what you added with the new not_staff policy.

This bug was because managers could still see the index page of all offices, but then couldn't access the individual offices on the list

novotnyjakub:
Maybe verify this with Jiten, but the problem was that managers could both list and see all offices. You removed the index permission, but they can still see each office if they change the url. I think that the view permission has to be something like this:

```ruby
def show?
  admin_or_mi? || (staff_or_manager? && same_office?)
end
```

novotnyjakub:
:+1: good to go!

PR #579, opened 2016-02-01
Prevent deleted user errors

colinbruce:
Show warning message when deleted user is invited

This should prevent the `PG::UniqueViolation: ERROR:  duplicate key
value violates unique constraint "index_users_on_email"` sentry warnings.

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -14,11 +14,9 @@ module Users
     end
 
     def create
-      user_for_authorisation
       authorize user_for_authorisation
-      if user_for_authorisation.deleted?
+      if @user.deleted?
         flash[:alert] = t('devise.invitations.user_exists', email: Settings.mail.tech_support)
-        @user = user_for_authorisation
         render :new
       else
         super
@@ -40,7 +38,7 @@ module Users
     end
 
     def user_for_authorisation
-      deleted_user_exists? || User.new(invite_params)
+      @user ||= deleted_user_exists? || User.new(invite_params)
     end
 
     def deleted_user_exists?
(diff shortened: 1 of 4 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -14,10 +14,15 @@ def new
     end
 
     def create
-      user_for_authorisation = User.new(invite_params)
+      user_for_authorisation
```
novotnyjakub: Does this have to be called separately, if it's passed to the method bellow?

app/controllers/users/invitations_controller.rb
```diff
@@ -14,10 +14,15 @@ def new
...
       authorize user_for_authorisation
-
-      super
+      if user_for_authorisation.deleted?
+        flash[:alert] = t('devise.invitations.user_exists', email: Settings.mail.tech_support)
+        @user = user_for_authorisation
+        render :new
```
novotnyjakub: I think rather than rendering the `:new` template, we should redirect back to the `/new` action, to prevent re-submission?

app/controllers/users/invitations_controller.rb
```diff
@@ -33,5 +38,13 @@ def build_role_list
     def invite_params
       params.require(:user).permit(:email, :role, :name, :office_id)
     end
+
+    def user_for_authorisation
+      deleted_user_exists? || User.new(invite_params)
```
novotnyjakub: You could cache the user `@user ||= ...` as it's called multiple times in the controller action.

novotnyjakub:
Very useful feature! And not actually as difficult as I thought it could get :)

app/controllers/users/invitations_controller.rb
novotnyjakub: OK, I think I understand now, because they could just choose a different e-mail and create the user again. But then it more seems like a validation concern, which could be displayed as an error on the `email` field? Probably something to ask Adrian about, from product perspective.

novotnyjakub:
:+1:

PR #580, opened 2016-02-01
Don't raise error on date_fee_paid validation

novotnyjakub:
The field is now not validated until `date_received` is a valid date. 

The error message on `date_received` field is not ideal, because if the format is incorrect, it simply says "The application must have been made in the last 3 months". We should create a new card to improve that message in case of an incorrect date. 

The `date_validator` would work well in case of being used with `ActiveRecord`, which provides `...before_type_cast` methods. But because we're only using `ActiveModel` and `Virtus`, that gem doesn't work as expected. But I think we can achieve similar thing maybe using a **custom Virtus coercer** or so.

### DIFF
--- a/app/models/forms/application/detail.rb
+++ b/app/models/forms/application/detail.rb
@@ -41,7 +41,7 @@ module Forms
         }
       end
 
-      with_options if: :refund? do
+      with_options if: :validate_date_fee_paid? do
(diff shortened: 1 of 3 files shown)


colinbruce:
LGTM :+1:

dotemacs:
Merge when ready.

PR #583, opened 2016-02-03
Update evidence accuracy error handling

colinbruce:
When the user submitted the evidence/accuracy page without
checking either checkbox, the boxes were highlighted in red,
but no error was shown.
screen shot 2016-02-03 at 10 17 57
This has added an error message to the view and set the text in the
locale file.
screen shot 2016-02-03 at 10 19 39


novotnyjakub:
:+1: all good, merge!

PR #589, opened 2016-02-09
Update locale handling

colinbruce:
Content have requested that the default submit buttons should lower case the model names.

This PR adds an I18n extension that allows us to lower case models within the YML file, instead of having to manually set the text on every submit button on every form.


dotemacs:
Go ahead.

PR #591, opened 2016-02-10
Created application redirect service

colinbruce:
When a user has clicked 'completed processing' on the summary page,
any attempts to view the individual process views should be redirected
to the current state's end point, e.g. deleted, processed, awaiting
payment, awaiting evidence.

The controller calls the service class which returns the relevant path,
this is overridden for create and confirmation to allow new items to be
created(!) and the final page to be re-shown if users need to get the
letters from the view.


dotemacs:
Do it.

PR #594, opened 2016-02-11
Force application redirect after completion

colinbruce:
Because the pages were stored in the browser history the users could
still use the browser back button.

This was designed to set the application processed page to expire and
force a reload/redirect if trying to view in the same browser session.


dotemacs:
Looks wholesome :custard:

PR #595, opened 2016-02-12
Update the Business Entity seed script

colinbruce:
It now only creates a BEC entry from the previously created office_jurisdiction objects.

This will allow users to create applications on the demo environments without having to manually create BEC codes.


dotemacs:
:+1:

PR #596, opened 2016-02-15
Display emergency banner

colinbruce:
Because the DWP has remained down for 4 days...

Show a warning to users, this will need to be removed manually...

Or we could write a checker to see if it should be displayed automatically!


dotemacs:
Looks MVP.

PR #598, opened 2016-02-18
Finance report update

colinbruce:
Updated dates and scope of state for the finance report.

The scope was including all applications, not only those successfully completed

Also the date was not being set to midnight at either end of the day as was expected by the users

### DIFF
--- a/app/builders/finance_report_builder.rb
+++ b/app/builders/finance_report_builder.rb
@@ -12,8 +12,9 @@ class FinanceReportBuilder
                   benefit_count benefit_sum income_count income_sum]
 
   def initialize(start_date, end_date)
-    @date_from = Date.parse(start_date.to_s)
-    @date_to = Date.parse(end_date.to_s)
+    @date_from = Date.parse(start_date.to_s).to_datetime
+    end_date = Date.parse(end_date.to_s)
+    @date_to = DateTime.new(end_date.year, end_date.month, end_date.day, 23, 59, 59, 'GMT')
   end
 
   def to_csv
--- a/app/builders/finance_report_builder.rb
+++ b/app/builders/finance_report_builder.rb
@@ -13,8 +13,7 @@ class FinanceReportBuilder
 
   def initialize(start_date, end_date)
     @date_from = Date.parse(start_date.to_s).to_datetime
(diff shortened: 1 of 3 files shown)


app/builders/finance_report_builder.rb
```diff
@@ -12,8 +12,9 @@ class FinanceReportBuilder
                   benefit_count benefit_sum income_count income_sum]
 
   def initialize(start_date, end_date)
-    @date_from = Date.parse(start_date.to_s)
-    @date_to = Date.parse(end_date.to_s)
+    @date_from = Date.parse(start_date.to_s).to_datetime
+    end_date = Date.parse(end_date.to_s)
+    @date_to = DateTime.new(end_date.year, end_date.month, end_date.day, 23, 59, 59, 'GMT')
```
csutter: Think you could use `DateTime#end_of_day` instead of the manual timings

app/models/views/reports/finance_report_data_row.rb
```diff
@@ -72,7 +72,8 @@ def applications
           select(:decision).
           where("decision IN ('part', 'full')").
           where('decision_date BETWEEN :d1 AND :d2', d1: @date_from, d2: @date_to).
-          where(business_entity_id: @business_entity.id)
+          where(business_entity_id: @business_entity.id).
+          where(state: 3)
```
csutter: This should probably use `state: :processed` to avoid the magic number

csutter:
Can't quite comment on the business logic impact of this, but code wise just two minor stylistic comments – otherwise :+1:

app/builders/finance_report_builder.rb
```diff
@@ -12,8 +12,9 @@ class FinanceReportBuilder
                   benefit_count benefit_sum income_count income_sum]
 
   def initialize(start_date, end_date)
-    @date_from = Date.parse(start_date.to_s)
-    @date_to = Date.parse(end_date.to_s)
+    @date_from = Date.parse(start_date.to_s).to_datetime
```
dan-palmer: You could use the [`DateTime#parse`](http://ruby-doc.org/stdlib-2.3.0/libdoc/date/rdoc/DateTime.html#method-c-parse) method here.

PR #601, opened 2016-02-23
fixed spelling mistake

colinbruce:
User pointed out the mis-spelling of received on the label


leighmoney:
Grateful to eagle eyed users!

PR #608, opened 2016-03-07
Update rubocop

colinbruce:
Update rubocop to 37.2 and rubocop-rails to 1.4.

Then making rubocop happy.

This returns us to current on the two gems...


novotnyjakub:
:+1: Let's get this in asap, so we can re-base ongoing branches on it.

PR #609, opened 2016-03-07
Remove superceded personae tests

colinbruce:
These were set up as an experiment in auto generating test personae.

They failed and should have been removed before now!


novotnyjakub:
This looks great :+1: I'm not sure exactly about the reason, but deleting is the best, get it in!

PR #611, opened 2016-03-08
Updated gems

colinbruce:
This addresses  security fix in Rails 4.2.5.2, but also removes the hard coded versions of rails and sprockets that were fixed after the last security update


novotnyjakub:
This would be good to get in as well (the Rails upgrade mainly), but the tests are failing for a strange reason. I can help with investigation, if you want to.

PR #614, opened 2016-03-10
Create OnlineApplication and its factory

novotnyjakub:
The model only contains basic validation, based on the decided database
structure.


colinbruce:
Looks fine as a base, we can always add if and when we need!
Merge when green.

PR #615, opened 2016-03-10
Add the API endpoint for publicapp submissions

colinbruce:
Include routing and return value tests

This will become a root branch for other features to merge into.


novotnyjakub:
:+1: Cool, I think we should merge this now? I'd then add Foreman setup, so that the 2 apps can be started up to work together.

PR #616, opened 2016-03-11
Online applications editting

novotnyjakub:
Second page of the online applications flow, which is the edit page. Staff should add details to the existing data, before they can complete processing of it.


colinbruce:
Looks good! :+1:

PR #617, opened 2016-03-11
Fix validation of boolean fields

novotnyjakub:
It's an important for on the OnlineApplication model, which is currently
not being able to save records.

The problem is that boolean fields can't be validated using presence
validator, but only using an inclusion validator. Subsequently the spec
then can't use the given shoulda matcher, but only can test for
inability to assign nil value to those fields.


colinbruce:
Looks good, merge now to make the other PRs easier... :+1: /signing-off

PR #618, opened 2016-03-14
Online applications summary

novotnyjakub:
The staff can now proceed do the summary page.

The prototype design has slightly different values for some fields (for example `None` instead of `0` for `Number of children`), but it's consistent with the other summary pages. So when we work on making them the same, this will change too.


colinbruce:
Looks good :+1:

PR #620, opened 2016-03-16
Create online application

colinbruce:
This is the MVP for receiving data (currently JSON) from the public app.

It will create an online_application, generate (and add) a (HWF-) reference number and save it to the database.

Things that still need work: 
* confirm format of reference!
* test non-generation of whatever format is decided on
* expand return values when errors occur.

### DIFF
--- a/app/builders/online_application_builder.rb
+++ b/app/builders/online_application_builder.rb
@@ -1,0 +1,12 @@
+class OnlineApplicationBuilder
+
+  def initialize(object)
+    @source = object
+  end
+
+  def build
+    generator = HwfReferenceGenerator.new
+    @source.merge(generator.attributes)
+    OnlineApplication.new(@source)
+  end
+end
--- a/app/builders/online_application_builder.rb
+++ b/app/builders/online_application_builder.rb
@@ -6,7 +6,7 @@ class OnlineApplicationBuilder
 
   def build
     generator = HwfReferenceGenerator.new
-    @source.merge(generator.attributes)
+    @source.merge!(generator.attributes)
     OnlineApplication.new(@source)
   end
 end
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -10,9 +10,9 @@ module Api
     def create
       online_submission = OnlineApplicationBuilder.new(public_app_params).build
       if online_submission.save
-        render(json: { result: true, message: online_submission.reference }.to_json)
+        render(json: { result: true, message: online_submission.reference })
       else
-        render(json: { result: false, message: 'Could not save online_submission' }.to_json)
+        render(json: { result: false, message: 'Could not save online_submission' })
       end
     end
 
@@ -20,7 +20,7 @@ module Api
 
     # rubocop:disable MethodLength
     def public_app_params
-      params.require(:jwt).permit(
+      params.require(:online_application).permit(
         :married,
         :threshold_exceeded,
         :benefits,
--- a/app/controllers/api/submissions_controller.rb
+++ b/app/controllers/api/submissions_controller.rb
@@ -8,14 +8,44 @@ module Api
     before_action :authenticate
 
     def create
-      result = { result: true, message: public_app_params['jwt'] }
-      render(json: result.to_json)
+      online_submission = OnlineApplicationBuilder.new(public_app_params).build
+      if online_submission.save
+        render(json: { result: true, message: online_submission.reference }.to_json)
+      else
+        render(json: { result: false, message: 'Could not save online_submission' }.to_json)
+      end
     end
 
     private
 
+    # rubocop:disable MethodLength
     def public_app_params
-      params.permit(:jwt)
+      params.require(:jwt).permit(
+        :married,
+        :threshold_exceeded,
+        :benefits,
+        :children,
+        :income,
+        :refund,
+        :date_fee_paid,
+        :probate,
+        :deceased_name,
+        :date_of_death,
+        :case_number,
+        :form_name,
+        :ni_number,
+        :date_of_birth,
+        :title,
+        :first_name,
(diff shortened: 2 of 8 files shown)


app/controllers/api/submissions_controller.rb
```diff
@@ -8,14 +8,44 @@ class SubmissionsController < ApplicationController
     before_action :authenticate
 
     def create
-      result = { result: true, message: public_app_params['jwt'] }
-      render(json: result.to_json)
+      online_submission = OnlineApplicationBuilder.new(public_app_params).build
+      if online_submission.save
+        render(json: { result: true, message: online_submission.reference }.to_json)
```
novotnyjakub: I think you can remove the `.to_json` method call as it's called automatically when rendered. 

Later we should probably encapsulate the successful / unsuccessful response to a presenter. But no need now.

app/controllers/api/submissions_controller.rb
```diff
@@ -8,14 +8,44 @@ class SubmissionsController < ApplicationController
...
+        render(json: { result: false, message: 'Could not save online_submission' }.to_json)
+      end
     end
 
     private
 
+    # rubocop:disable MethodLength
     def public_app_params
-      params.permit(:jwt)
+      params.require(:jwt).permit(
```
novotnyjakub: This goes throughout the PR, but I was thinking about calling the parameter something more current, like 'online_application', or 'application', ... Just in case we don't get to do the jwt soon.

spec/builders/online_application_builder_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  describe '#build' do
+    subject(:build_submission) { described_class.new(submission).build }
+
+    it 'returns non persisted OnlineApplication' do
+      is_expected.to be_a(OnlineApplication)
+      is_expected.not_to be_persisted
+    end
+
+    describe 'email contact' do
```
novotnyjakub: I'm thinking that these are maybe not a responsibility of the builder? I think we should just store the data as it comes, assuming it's valid at this point.

spec/builders/online_application_builder_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      describe 'when applicant responds `no`' do
+        let(:submission) { attributes_for :public_app_submission, :email_contact }
+
+        it { is_expected.to eql 'foo@bar.com' }
+      end
+    end
+
+    describe 'children field' do
```
novotnyjakub: The same applies to here as above, the field will already be prepared on the public app side, so you'll only have a number in `children` field.

spec/builders/online_application_builder_spec.rb
novotnyjakub: I suppose the real sanitisation and conversion will happen when we move the data from `OnlineApplication` to `Application` and then we need to do these various conversions.

spec/services/hwf_reference_generator_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe HwfReferenceGenerator, type: :service do
```
novotnyjakub: I'm thinking that we should probably try to test the collision case as well. Perhaps if the actual reference number string was generated by a private method, let's say `reference_string`, then you can use RSpec stub to return different values each time:

```ruby
before do
  create(:online_submission, reference: 'collision')
  allow(ReferenceGenerator).to receive(:reference_string).and_return('collision', 'no-collision')
end
```

And afterwards test that it was the 'no-collision' winner. Something like that?

novotnyjakub: And it'd be really good cause later we can generalise this to be used for our other reference generator and then somehow merge to be the same or very close to be the same.

spec/controllers/api/submissions_controller_spec.rb
```diff
@@ -20,6 +22,13 @@
         it { is_expected.to include 'message' }
         it { is_expected.to include 'result' }
       end
+
+      describe 'when sent invalid data from the public' do
```
novotnyjakub: Reading this I realised that this controller spec is more of a feature test, rather than controller spec. And then I didn't find a feature test, so I looked into it a bit more. 

I think we definitely should have a feature (request) spec to handle various cases, which are end-to-end. Including authentication problems, input errors, etc. 
The controller spec is theoretically only concerned about the various objects to call, rather then inputs and outputs. Hope that makes sense. 

I think we can go with this, though and look into the request testing as a separate bit, if you want?

spec/builders/online_application_builder_spec.rb
CeeBeeUK: Noted, these were a holdover from when the transformation was happening there...

colinbruce: Noted, these were a holdover from when the transformation was happening there...

spec/controllers/api/submissions_controller_spec.rb
colinbruce: Agreed, happy to discuss testing strategy, as per [pivotal](https://www.pivotaltracker.com/story/show/112878143)

app/builders/online_application_builder.rb
```diff
@@ -0,0 +1,12 @@
+class OnlineApplicationBuilder
+
+  def initialize(object)
+    @source = object
+  end
+
+  def build
+    generator = HwfReferenceGenerator.new
+    @source.merge(generator.attributes)
```
novotnyjakub: I just ran an integration from the public app and discovered you have here "the missing bang classic" :-) I do this all the time!
The `@source.merge` does the merge, but returns it. We need `@source.merge!` to merge the reference to the `@source`.

novotnyjakub:
:+1: Should we merge this now?

PR #624, opened 2016-03-18
Monitor dwp

colinbruce:
Monitor the last 20 requests and return a state based on the number of 400 Bad Requests received

This is then called from the home controller to display the correct banner.

Update the healthcheck page to call the monitor too.

To do
===
- [x] rebase on master
- [x] front end review of new css
- [x] code review

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -6,6 +6,7 @@ class HomeController < ApplicationController
     load_graphs_for_admin
     load_waiting_applications
     @search_form = Forms::Search.new
+    @state = DwpMonitor.new.state
   end
 
   def search
@@ -16,6 +17,7 @@ class HomeController < ApplicationController
       redirect_to(edit_online_application_path(online_application))
     else
       load_waiting_applications
+      @state = DwpMonitor.new.state
       render :index
     end
   end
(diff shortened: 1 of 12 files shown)


app/controllers/home_controller.rb
```diff
@@ -6,6 +6,7 @@ def index
     load_graphs_for_admin
     load_waiting_applications
     @search_form = Forms::Search.new
+    @state = DwpMonitor.new.state
```
novotnyjakub: Can you add this assignment to the controller spec as there's no other place we test that it's there?

app/views/shared/dwp/_online.html.slim
```diff
@@ -0,0 +1,3 @@
+.row.pad-top-thirty-px
```
novotnyjakub: I'm not sure if we want to display the green box there permanently, when the DWP check is available? It rings a bell that I overheard that sort of conversation some time ago, but I'm not sure about it.

colinbruce: I agree, this was a first pass at the problem to alert the users (and us, via the healthcheck) that the status had changed.  My long term plan has a store for the time of state changes so that we could turn off the green bar after X hours.

app/controllers/home_controller.rb
colinbruce: Updated...

axemonkey:
Front-endy stuff looks :+1: to me.

PR #627, opened 2016-03-22
Add feedback_opt_in to online_applications

colinbruce:
This aligns the staff app with the public app and ensures that the new feedback_opt_in field is recorded


novotnyjakub:
:+1: nice and clean!

PR #628, opened 2016-03-23
Complete online application

novotnyjakub:
This finishes the online applications journey by allowing the staff to complete processing. They should get relevant result and the same confirmation page as paper based applications.


colinbruce:
gladiator

PR #629, opened 2016-03-24
Copy reference from OnlineApplication

novotnyjakub:
We wanted to preserve the original reference, but it was forgotten to be
copied. So this is a bug fix for that feature.


colinbruce:
Looks good, merge when you want.  I'm not sure if you want to get this in before the demo?

novotnyjakub:
Thanks! We'll merge after the demo.

PR #631, opened 2016-03-30
Email online completion

colinbruce:
When the staff user processes an online application where the public user has provided an email address,  this will send an email to them!

Still needed
===
- [x] Inline styling work on the email
- [ ] code review


axemonkey:
I think this should be enough for the HTML email layout.

novotnyjakub:
:+1: looks good, let's get that in. 

One thing which starts to confuse me is the terminology. I'm not sure `OnlineApplication` is the right thing as everything is "online". Also `OnlineMailer` as a result is also not self explanatory. I don't think we have to change anything now, but maybe think about different naming? In the team, we call it "Public submission" so maybe that would be closer?

PR #632, opened 2016-03-31
Remove reference feature toggle

colinbruce:
From April 1st 2016, we don't need conditional display of reference numbers.

This removes the helper and it's tests, updates the features to remove explicit before/after date tests and leaves only those to check that is is displayed and that references to the remission register are removed.


novotnyjakub:
:+1: brilliant!

PR #633, opened 2016-04-01
Lock PhantomJS to 1.9.8

colinbruce:
The Travis CI build pre-installs PhantomJS-1.9.8, but Jasmine was trying to use 2.1.1.0. 

This meant that Travis needed to download and install it each time travis ran... this often failed :-1:

I've locked the Gemfile to use Travis' pre-installed version, it now.just.works. :boom:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -176,7 +176,7 @@ GEM
     ice_nine (0.11.1)
     jasmine (2.4.0)
       jasmine-core (~> 2.4)
-      phantomjs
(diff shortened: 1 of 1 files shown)


novotnyjakub:
Is this gonna be ok for @axemonkey in terms of using older `phantomjs`?

axemonkey:
Right now we don't run any FE tests on CI, so it won't make any difference. 😕

novotnyjakub:
Ok, let's put it on the backlog to start doing that :)

PR #636, opened 2016-04-04
Search update

colinbruce:
Minor tweaks for staff searches
* upper case the search string (prevents hwf-abc-123 not found, when HWF-ABC-123 exists)
* turn off auto-complete on the search field

[finishes #116868539]


novotnyjakub:
:+1:

PR #637, opened 2016-04-04
Remove foundation

axemonkey:
The staff app originally got built on top of prototype code that Kellie made with the help of Zurb Foundation. That's fine for prototyping, but it's not what we want an MOJ application to look like, even if it's staff-facing. This huge PR is to remove Foundation and style all pages using the GOV.UK elements stlye and grid layout.

Before
---
screen shot 2016-04-11 at 11 30 43

After
---
screen shot 2016-04-11 at 11 31 03


Gemfile
```diff
@@ -29,7 +29,11 @@ gem 'coffee-rails', '~> 4.1.0'
 # gem 'therubyracer', platforms: :ruby
 
 # use GDS assets, styles etc...
-gem 'govuk_frontend_toolkit', '~> 4.2.1'
+gem 'govuk_frontend_toolkit',
+  git: "https://github.com/alphagov/govuk_frontend_toolkit_gem.git", submodules: true
```
novotnyjakub: I'm slightly concerned about using `master` versions from Git. I'll have a look if we should try to peg it to a version or what are the alternatives.

app/assets/stylesheets/local/alerts.scss
```diff
@@ -0,0 +1,38 @@
+.alert {
```
novotnyjakub: Maybe if we have a bit of a slack in the next sprint, we can speak to some other devs in the department and try to back port this to the `moj_template`.

app/views/layouts/application.html.slim
```diff
@@ -1,48 +1,50 @@
...
+            -if current_user.elevated?
+              li= link_to 'View office', current_user.office
+              li= link_to 'View staff', users_path
+            li= link_to 'Sign out', destroy_user_session_path, :method => :delete
+
+
+- content_for :javascripts
+  = javascript_include_tag('application.js')
+
+= render template: "layouts/local-template"
```
novotnyjakub: Have you looked at using the layout from `layouts/moj_template`, which is part of the `moj_template`?

novotnyjakub:
So it wasn't as painful as I thought it could get :) Few things I really like:

* a lot of deletions (of already existing css in the gems), so maintainability very much increased
* cleaner HTML, because of the more concise css classes
* not too many spec changes, so it's relatively save to merge

It would be good to keep an eye on the upcoming form builder that some devs are working on, which would improve this even more!

app/views/layouts/application.html.slim
axemonkey: Absolutely, that's where we started. But that doesn't allow for a custom footer or a menu in the `proposition` area. I think `moj_template` may actually not be as flexible as the `govuk_template`. Either way, to make Kellie's design work in `moj_template` meant having a local copy we could override and hack about.

Gemfile
axemonkey: I had hell trying to get all the govuk gems and moj_template to actually work. I had to manually hack `Gemfile.lock` in the end.

novotnyjakub:
:+1: ok, let's do it!

PR #638, opened 2016-04-04
Fix and update reference keys (new PR)

novotnyjakub:
This is a preparation for the trial -> production migration. We need to be able to update all primary keys and cascade those changes to all referenced tables. This PR adds missing foreign key database constraints. And also makes sure that all the foreign keys have UPDATE ON CASCADE setup.


colinbruce:
Looks good, if we aren't running the rake task, should we push it to a future PR?

novotnyjakub:
Probably a good idea :) I've just pushed a commit which deletes it for now..

PR #639, opened 2016-04-04
Update email functions

colinbruce:
Update the email functionality:
* add a refund email
* update content of both refund and confirmation emails

This allows users to receive a refund email correctly.
This will also allow Leigh to manage the email content directly.

Todo
===
- [x] Deliver [Ticket #116870885](https://www.pivotaltracker.com/story/show/116870885)
- [x] Review FE changes 
- [x] Code review


novotnyjakub:
The code looks good, can be merged. I'm only slightly concerned about the naming convention - confirmation vs. refund. Because both the scenarios are confirmation, just 2 different types.

colinbruce:
Good point - I was planning on renaming the mailers, but didn't want to trip over Clive's work.

novotnyjakub:
That makes sense.. Let's get it in then.

novotnyjakub:
:+1: yep, I like the new naming better. Let's get it in and test!

PR #641, opened 2016-04-08
Support for trial -> production migration

novotnyjakub:
For the moment it's rake tasks to support the migration, but it may be more if needed.


colinbruce:
Looks good so far, good luck deploying them :)

PR #649, opened 2016-04-18
Restore dwp banner

colinbruce:
This restores the conditional display of the DWP status banner.

It got removed during the foundation removal.

Tests have been added to ensure that the controller generates the correct state and that the view displays the correct text.

### DIFF
--- a/spec/support/dwp_setup.rb
+++ b/spec/support/dwp_setup.rb
@@ -1,0 +1,25 @@
+module DwpSetup
+
+  def build_with_bad_requests(yes_response = 5, bad_requests = 5)
+    teardown
+    create_list :benefit_check, yes_response, :yes_result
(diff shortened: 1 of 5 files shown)


novotnyjakub:
👍  looks alright.

spec/support/dwp_setup.rb
```diff
@@ -0,0 +1,25 @@
+module DwpSetup
+
+  def build_with_bad_requests(yes_response = 5, bad_requests = 5)
```
novotnyjakub: I'd maybe consider prefixing these methods with `dwp_checker` because in the code they're used without the module. But it's not a big issue.

colinbruce: Oops.  My web page didn't refresh until I merged...

PR #651, opened 2016-04-18
Cleanup rake tasks

novotnyjakub:
This removes the no longer needed rake tasks for trial migration and BEC import. 

The state migration classes in `lib` couldn't be removed just yet, because they're using in existing migrations. We would need to squeeze old migrations (which we probably should) and then we can remove them.


colinbruce:
👍

PR #653, opened 2016-04-19
Allow justice.gsi.gov.uk users to register

novotnyjakub:
I decided to just go ahead with the existing regexp solution. I started making it more generic and configurable, but then I decided not to pollute the codebase with unnecessary complexity.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,8 +19,7 @@ class User < ActiveRecord::Base
 
   scope :by_office, ->(office_id) { where('office_id = ?', office_id) }
 
-  email_regex =
-    /\A([^@\s]+)@(justice\.gsi|hmcourts-service\.gsi|hmcts\.gsi|digital\.justice)\.gov\.uk\z/i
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -19,7 +19,7 @@ class User < ActiveRecord::Base
 
   scope :by_office, ->(office_id) { where('office_id = ?', office_id) }
 
-  email_regex = /\A([^@\s]+)@(hmcourts-service\.gsi|hmcts\.gsi|digital\.justice)\.gov\.uk\z/i
+  email_regex = /\A([^@\s]+)@(justice\.gsi|hmcourts-service\.gsi|hmcts\.gsi|digital\.justice)\.gov\.uk\z/i
```
colinbruce: Not a regex expert, but would
```ruby
  email_regex = /\A([^@\s]+)@((justice|hmcourts-service|hmcts)\.gsi|digital\.justice)\.gov\.uk\z/i
```
be better, save repeating the .gsi each time?

novotnyjakub: Good point, let me test it!

colinbruce:
Looks good 👍

PR #655, opened 2016-04-22
Display DWP maintenance banner for date range

novotnyjakub:
A temporary banner timed to be displayed on the homepage.

The view spec is a bit hacky, but the view has to be extended to include the `dwp_maintenance?` method, which then can be stubbed.


colinbruce:
👍

PR #656, opened 2016-04-28
Update HWF reference generator

colinbruce:
Replace hex references with an extended character set with no
ambiguities. "3 4 6 7 9 A C D E F G H J K L M N P Q R T V W X Y"

The new algorithm delivers collisions at about 1:10,000

Created | Time taken (sec) | Collisions | Percent
:-----: | ---------: | :--------: | :-----:
10,000 | 0.04 | 1 | 0.01%
100,000 | 0.4 | 21 | 0.02%
1,000,000 | 5.1 | 2080 | 0.21%
7,000,000 | 43.0 | 98586 | 1.41%

Obviously, that will increase over time but it seems acceptable for now.

Algorithm based on speed tests done [here](http://goo.gl/QwVedv) and [here](http://codereview.stackexchange.com/a/15997)

### DIFF
--- a/app/services/hwf_reference_generator.rb
+++ b/app/services/hwf_reference_generator.rb
@@ -1,5 +1,8 @@
 class HwfReferenceGenerator
 
+  DICTIONARY = %w[ 3 4 6 7 9 A C D E F G H J K L M N P Q R T V W X Y ].freeze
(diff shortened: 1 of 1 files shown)


novotnyjakub:
👍

PR #657, opened 2016-04-29
Enable default jurisdiction

colinbruce:
When processing online-applications a users' default jurisdiction
was not being set on the new object and being shown on the view

[Pivotal](https://www.pivotaltracker.com/story/show/118470941)

[finishes #118470941]

### DIFF
--- a/app/controllers/online_applications_controller.rb
+++ b/app/controllers/online_applications_controller.rb
@@ -4,6 +4,7 @@ class OnlineApplicationsController < ApplicationController
   def edit
     authorize online_application
     @form = Forms::OnlineApplication.new(online_application)
+    @form.update_attributes(jurisdiction_id: current_user.jurisdiction_id)
     assign_jurisdictions
   end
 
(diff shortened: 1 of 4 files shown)


colinbruce:
@novotnyjakub Can you have a look at this? I'm happy with the code but the tests seem.. flakey! :/

app/controllers/online_applications_controller.rb
```diff
@@ -4,6 +4,7 @@ class OnlineApplicationsController < ApplicationController
   def edit
     authorize online_application
     @form = Forms::OnlineApplication.new(online_application)
+    @form.update_attributes(jurisdiction_id: current_user.jurisdiction_id)
```
novotnyjakub: I would personally try to wrap this functionality somewhere else, similarly like we have the `ApplicationBuilder` for `Application` (even though of course that's an actual builder and more complex). It looks like logic the controller should not worry about. 

Perhaps we could have `assign_default_jurisdiction` on the `OnlineApplication` which accepts `current_user` and does it there. That would simplify testing that particular piece (cause it would be in the model). And in the controller test bellow, you can just check that method has been called on the model.

novotnyjakub:
👍  good to go!

PR #658, opened 2016-05-03
Prevent public HWF reference reuse

colinbruce:
When a member of staff searched for a HWF reference that had already been processed, they would be allowed to continue until they hit complete processing, at which point it would crash.

This PR shows an error if the number has previously been processed.

If it was by the users office, it includes a routed link.
If it was by a different office, it includes the name of the office that processed it, but no link.

This is because of the Pundit policy for accessing applications for different offices.

[Pivotal](https://www.pivotaltracker.com/story/show/118470941)

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -11,7 +11,6 @@ class HomeController < ApplicationController
 
   def search
     @search_form = Forms::Search.new(search_params)
-
     online_application = search_and_return
     if online_application
       redirect_to(edit_online_application_path(online_application))
@@ -73,12 +72,18 @@ class HomeController < ApplicationController
 
   def search_and_return
     if @search_form.valid?
-      begin
-        OnlineApplication.find_by!(reference: @search_form.reference.upcase)
-      rescue ActiveRecord::RecordNotFound
-        @search_form.errors.add(:reference, :not_found)
-        nil
+      @search = ApplicationSearch.new(@search_form.reference.upcase, current_user)
+      @matched = @search.for_hwf
+      if @matched.is_a? OnlineApplication
+        return @matched
+      else
+        @search_form.errors.add(:reference, @search.error_message)
       end
+      return nil
     end
+  end
+
+  def can_show(application)
+    application.office == current_user.office
   end
 end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -72,18 +72,14 @@ class HomeController < ApplicationController
 
   def search_and_return
     if @search_form.valid?
-      @search = ApplicationSearch.new(@search_form.reference.upcase, current_user)
-      @matched = @search.for_hwf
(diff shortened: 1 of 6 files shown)


leighmoney:
@colinbruce - have tweaked those errors a little.

app/services/application_search.rb
```diff
@@ -0,0 +1,57 @@
+class ApplicationSearch
+
+  attr_reader :error_message
+
+  def initialize(reference, current_user)
```
novotnyjakub: I would suggest looking at the option of not passing `current_user` here, but the `ApplicationPolicy` instead. Because bellow you're testing if the office matches, but this condition can change by the policy and this could bypass it.

If you want I can have a look at how to use Pundit this way to help..

app/controllers/home_controller.rb
```diff
@@ -73,12 +72,18 @@ def search_params
 
   def search_and_return
     if @search_form.valid?
-      begin
-        OnlineApplication.find_by!(reference: @search_form.reference.upcase)
-      rescue ActiveRecord::RecordNotFound
-        @search_form.errors.add(:reference, :not_found)
-        nil
+      @search = ApplicationSearch.new(@search_form.reference.upcase, current_user)
```
novotnyjakub: The `@search` here could be just a local variable, it's not exposed to the view.

Also I think it's cleaner if you `upcase` the reference inside the service, it keeps the requirements on the interface cleaner.

app/controllers/home_controller.rb
```diff
@@ -73,12 +72,18 @@ def search_params
...
+      if @matched.is_a? OnlineApplication
+        return @matched
+      else
+        @search_form.errors.add(:reference, @search.error_message)
       end
+      return nil
     end
   end
+
+  def can_show(application)
```
novotnyjakub: This is not used anywhere, is it? If yes, same as bellow, we should use the `ApplicationPolicy`.

colinbruce: That got moved into ApplicationSearch and should have been deleted! :)

app/services/application_search.rb
```diff
@@ -0,0 +1,58 @@
...
+      @error_message = I18n.t(:processed_by, scope: scope, office_name: application_office)
+    end
+  end
+
+  def application_exists
+    @application ||= Application.find_by(reference: @reference.upcase)
+  end
+
+  def user_can_access
+    ApplicationPolicy.new(@current_user, @application).show?
```
novotnyjakub: This is now exactly what it says 👍  Only one little detail - there's a recommended way of retrieving policies outside views and controllers: https://github.com/elabs/pundit#manually-retrieving-policies-and-scopes. It's just to be backwards compatible if they make changes.

novotnyjakub:
Apart from my last comment about retrieving the policy using `Pundit.policy`, it's good to go! 👍

novotnyjakub:
👍  perfect

PR #659, opened 2016-05-04
Trial dashboard

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/117995369)

Surface data from public submissions to admin users.

Display total completed all time + last 7 days
Display processed by court, all time + last 7 days
screen shot 2016-05-04 at 12 49 42

Work in progress
===
Time taken to between online submission and application processed
screen shot 2016-05-04 at 17 01 07

### DIFF
--- a/spec/controllers/reports_controller_spec.rb
+++ b/spec/controllers/reports_controller_spec.rb
@@ -66,5 +66,15 @@ RSpec.describe ReportsController, type: :controller do
         expect(assigns(:report_data).count).to eql(1)
       end
     end
+
+    describe 'GET #public' do
+      before { get :public }
+
+      subject { response }
+
+      it { is_expected.to have_http_status(:success) }
+
+      it { is_expected.to render_template :public }
+    end
   end
 end
--- a/spec/models/views/reports/public_submission_data_spec.rb
+++ b/spec/models/views/reports/public_submission_data_spec.rb
@@ -40,4 +40,26 @@ RSpec.describe Views::Reports::PublicSubmissionData do
 
     it { is_expected.to be_a Array }
   end
+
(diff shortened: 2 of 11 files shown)


spec/controllers/reports_controller_spec.rb
```diff
@@ -66,5 +66,15 @@
         expect(assigns(:report_data).count).to eql(1)
       end
     end
+
+    describe 'GET #public' do
```
novotnyjakub: I wonder if we should test that the `@data` variable is assigned. 

I usually use double to verify the right object is being assigned, but I'm not saying it's the best option:

```ruby
let(:report_view) { double } 

before do
  allow(Views::Reports::PublicSubmissionData).to receive(:new).and_return(report_view)
end

it 'assigns the report view as @data' do
  expect(assigns(:data)).to eql(report_view)
end
```

Something between those lines.

spec/models/views/reports/public_submission_data_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require 'rails_helper'
+
+RSpec.describe Views::Reports::PublicSubmissionData do
+
+  subject(:data) { described_class.new }
+
+  describe '#submission_all_time_total' do
+    subject { data.submission_all_time_total }
+
+    it { is_expected.to be_a Integer }
```
novotnyjakub: I suppose this is about importance (how important it is if the report wasn't working correctly). But I'd test some specific values, read from the DB. You can use factories to prime it with the specific setup and test your queries are working. Let's discuss it tomorrow, though.

novotnyjakub:
👍  looks good, let's get that in

PR #663, opened 2016-05-17
nasty css hacky border rubbish for fee box

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/118280569)

The border on the fee box is faked using a wrapping `div` so that it encompasses the currency symbol. This broke badly when the error handling styles were invoked. This works around that with further fakey bordery nonsense.

Before
---
screen shot 2016-05-17 at 14 50 01

After
---
screen shot 2016-05-17 at 14 49 49


novotnyjakub:
👍 let's get it in!

PR #666, opened 2016-05-20
Performance platform integration

colinbruce:
Add Google Analytics and Google Tag Manager to the staff app to enable generation of GDS performance Platform metrics.

Use Tags, Triggers and events to generate completed events upon form submission.

Initially, I have added the google JS code snippets and I will try configuring the recording of events between Google Tags and Google Analytics events.

### DIFF
--- a/app/views/layouts/local-template.html.erb
+++ b/app/views/layouts/local-template.html.erb
@@ -61,7 +61,7 @@
   </head>
 
   <body class="<%= content_for?(:body_classes) ? raw("#{yield(:body_classes)}") : "#{config_item(:phase).to_s.downcase} #{config_item(:product_type).to_s.downcase}" %>">
-    <%= render :partial => 'shared/tag_manager' %>
+    <%= render 'shared/tag_manager' %>
     <script type="text/javascript">document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 
     <%= yield :body_start %>
@@ -132,7 +132,7 @@
         <% end %>
 
 
-    <%= render :partial => 'shared/footer' %>
+    <%= render 'shared/footer' %>
 
     <!--end footer-->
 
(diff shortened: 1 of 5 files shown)


app/views/layouts/local-template.html.erb
```diff
@@ -61,6 +61,7 @@
   </head>
 
   <body class="<%= content_for?(:body_classes) ? raw("#{yield(:body_classes)}") : "#{config_item(:phase).to_s.downcase} #{config_item(:product_type).to_s.downcase}" %>">
+    <%= render :partial => 'shared/tag_manager' %>
```
novotnyjakub: A code style improvement: `<%= render 'shared/tag_manager' %>` - the hash is not necessary.

novotnyjakub:
👍  code-wise it's ok apart from the little comment above. But I don't know much about Google Tag manager. Is it necessary for this scenario?

app/views/layouts/local-template.html.erb
colinbruce: Noted, I'll push an update.  I wasn't going to, but I saw the footer used that structure at line 135 and thought I'd keep it the same for consistency

colinbruce:
I wrote a long description explaining why I used it, but refreshed the page and somehow lost it! 😞 

Short-ish reason is:-
Google analytics, when calling "ga('send', 'event',  ...)", fires a GET event.  If that is called from a form.onSubmit action it will _sometimes_ work.  But by no means always.  ONe report was a 50-60% success rate, as the form's POST swamps the GA GET.

Using tag manager, allegedly, allows us to monitor a form submission from a certain page and Tag Manager then fires the GA send event.  After validation has passed.

Pro : Can be updated and managed outside the code-base
Con: It could be a bit obscure for maintaining

novotnyjakub:
Cool, I was particularly worried about the ongoing maintenance. So as long as we document it on Confluence or somewhere, it should be fine. I'll ask you later to give me a quick overview of Google Tag Manager.

PR #668, opened 2016-05-25
Updated slim layout for JS

colinbruce:
A cut and paste error left issues with the tag manager JS block...


novotnyjakub:
I didn't do a very good job reviewing this before... Now it seems like a correct conversion to slim. 

I'm just thinking, why is the partial `tag_manager` included right after the body, rather then together with all the scripts at the bottom of the page? Is it a recommendation from Google, or could we move it down?

colinbruce:
It's a Google directive:

> Install Google Tag Manager
> Copy the code below and paste it on to every page of your website. Place it immediately after the opening <body> tag.

novotnyjakub:
👍  ok, cool, just wanted to make sure we don't put it there unnecessarily. Let's merge then!

PR #670, opened 2016-06-01
Code climate update

colinbruce:
Code climate was complaining about an insecure git reference, also added the suggested codeclimate.yml files to allow better analysis and integrations


novotnyjakub:
It's annoying that Bundler doesn't automatically use ssl. Even though I thought the git protocol was secure. That's a good fix 👍 

I'm not sure about all the `.eslintrc` configuration, because I don't know anything about it. But if it's useful, let's get it in!

👍

colinbruce:
I just want that 4.0 back! :)

The linters were auto generated by code-climate, it's essentially what they run remotely, but this allows more granular control.  They say! :)

novotnyjakub:
Cool, sounds good then!

PR #671, opened 2016-06-02
Processed application display

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/120572955) to update the processed application display to
- [x] Correct the `Error` display - [original pivotal error ticket](https://www.pivotaltracker.com/story/show/119828519)
- [x] Add the processing summary - [design](https://dsdmoj.atlassian.net/wiki/pages/viewpage.action?pageId=52920456)

New display of `return`ed applications, includes processing summary:
image

### DIFF
--- a/app/models/views/application_overview.rb
+++ b/app/models/views/application_overview.rb
@@ -94,7 +94,7 @@ module Views
       @application.reference if evidence_check_or_part_payment?
     end
 
-    def decision_type
+    def return_type
       {
         'evidence_check' => 'evidence',
         'part_payment' => 'payment'
--- a/app/models/views/application_overview.rb
+++ b/app/models/views/application_overview.rb
@@ -94,6 +94,10 @@ module Views
       @application.reference if evidence_check_or_part_payment?
     end
 
+    def decision_type
+      @application.decision_type.humanize.downcase if @application.decision_type
+    end
+
     private
 
     def evidence_check_or_part_payment?
--- a/app/models/views/application_overview.rb
+++ b/app/models/views/application_overview.rb
@@ -95,7 +95,10 @@ module Views
     end
 
     def decision_type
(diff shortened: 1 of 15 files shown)


GitCop:

 * Commit: d2472b9b755f3fd11974da6fc818b4d4360004c0

---

joshlow:
Possible copy for error message:

Not eligible because evidence or payment not provided

app/models/views/application_overview.rb
```diff
@@ -94,6 +94,13 @@ def reference
       @application.reference if evidence_check_or_part_payment?
     end
 
+    def decision_type
+      {
+        'evidence_check' => 'evidence',
+        'part_payment' => 'payment'
```
novotnyjakub: What if the `decision_type` is an application? It's not needed in the overview?

app/models/views/processed_data.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def initialize(application)
+      @application = application
+    end
+
+    def application_processed
+      build_return_hash @application
+    end
+
+    def application_deleted
```
novotnyjakub: I would follow the same pattern here as the other public methods - to be one liners which delegate the hash creation to a private method.

novotnyjakub:
I think it's very neat and clean! Couple of small comments, but nothing major.

app/models/views/application_overview.rb
colinbruce: It shouldn't be... there is no `return` function for applications is there?

novotnyjakub: No, there isn't. But the class is an `ApplicationOverview` and the method is called `decision_type`, which doesn't suggest you can't use it for an `Application` object itself, if that makes sense.

novotnyjakub:
👍  let's go!

PR #673, opened 2016-06-03
Handling DWP checks for out of time refunds

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/116873277)

We cause errors when we ask the DWP for checks longer than 3 months ago, so when a refund is processed from more than three months ago.  Don't even send a check to the DWP and render a pertinent error message to to the user.

### To do
- [x] Don't send DWP request if date > 3 months old
- [x] Display situationally appropriate error message to the user
- [ ] @joshlow to review error message

### DIFF
--- a/app/controllers/benefit_overrides_controller.rb
+++ b/app/controllers/benefit_overrides_controller.rb
@@ -35,11 +35,19 @@ class BenefitOverridesController < ApplicationController
   end
 
   helper_method def error_message_partial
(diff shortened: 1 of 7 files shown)


colinbruce:
@joshlow The new message is [here](https://github.com/ministryofjustice/fr-staffapp/blob/d71ca4efac148f3883a18e09aa89633d7733dfed/config/locales/en.yml#L47)
and is displayed as 
image

Happy to discuss on Monday AM.

novotnyjakub:
👍  looks good!

PR #676, opened 2016-06-16
Update savings on staffapp

colinbruce:
Implement an improved savings flow for staff processing paper applications
===
[This ticket](https://www.pivotaltracker.com/story/show/117992675) improves the on screen experience to better match the paper form.

Still to do
===
- [x] Add tests for the WIP commit
- [x] Write a proper PR description
- [x] Refactor the pass/fail service
- [x] Improve the JS handling on the view
- [x] Check validation message changes
- [x] Mark as ready for review
- [ ] Merge in all subsequent PR's


app/models/forms/application/savings_investment.rb
```diff
@@ -6,17 +6,18 @@ class SavingsInvestment < ::FormObject
 
       def self.permitted_attributes
         {
-          threshold_exceeded: Boolean,
-          partner_over_61: Boolean,
-          high_threshold_exceeded: Boolean
+          min_threshold_exceeded: Boolean,
+          over_61: Boolean,
+          max_threshold_exceeded: Boolean,
+          amount: Decimal
```
novotnyjakub: In the public app I assumed we're not dealing with pence, only pounds, so I used `Integer`. I'll try to find out what's correct.

app/models/saving.rb
```diff
@@ -0,0 +1,4 @@
+class Saving < ActiveRecord::Base
+  belongs_to :application, required: true, inverse_of: :saving
+  belongs_to :jurisdiction
```
novotnyjakub: I don't think this belongs here :)

app/services/partner_age_check.rb
```diff
@@ -1,34 +0,0 @@
-# TODO: Make a validator out of this service
-class PartnerAgeCheck
```
novotnyjakub: 👍

app/services/savings_pass_fail_service.rb
```diff
@@ -0,0 +1,67 @@
+class SavingsPassFailService
+
+  def initialize(saving)
+    @saving = saving
+  end
+
+  def calculate!
+    @saving.fee_threshold = FeeThreshold.new(@saving.application.fee).band
+    @saving.passed = calculate_pass_fail
+    unless @saving.passed
```
novotnyjakub: I would maybe abstract this to a method, to make the `calculate!` more readable. Something like `set_application_outcome_when_failed` or so.

app/services/savings_pass_fail_service.rb
```diff
@@ -0,0 +1,67 @@
+class SavingsPassFailService
```
novotnyjakub: I don't see anywhere the min and max thresholds being stored on the `Saving` object. It's done in the migration for the existing applications, but not for the new ones. Unless I missed something?

spec/models/views/overview/savings_and_investments_spec.rb
```diff
@@ -3,62 +3,62 @@
...
-
-    context 'when user selected "no" to threshold_exceeded' do
-      let(:application) { build_stubbed :married_applicant_under_61, threshold_exceeded: false }
-
-      it { is_expected.to eq nil }
-    end
+    it { is_expected.to eql '£3500' }
   end
+  #
+  # describe '#combined_savings_valid?' do
```
novotnyjakub: Let's remove this commented stuff, if it's not needed.

app/services/savings_pass_fail_service.rb
CeeBeeUK: They are saved in the form object, before this service is called

colinbruce: They are saved in the form object, before this service is called

novotnyjakub: They don't seem to be: 

```ruby
       def fields_to_update
          {
           min_threshold_exceeded: min_threshold_exceeded,
           over_61: over_61,
           max_threshold_exceeded: max_threshold_exceeded,
           amount: amount
          }
        end
```

colinbruce: You, sir, are correct!  Two many similarly named fields!   Yes, I was going to add a setting for min and max values and import them... 
On my todo list now! :)

novotnyjakub: Yeah, I missed it originally as well, when scanning through it the first time.

novotnyjakub:
@colinbruce one important thing I realised is that we can't merge this until we adjust the public submission endpoint.. Let's decide how to do that on Monday..

axemonkey:
JS fixed @colinbruce @novotnyjakub 👍

joshlow:
@colinbruce - copy suggestions for error messages

Please answer the savings question

Please enter the amount of savings and investments

Please confirm if the partner is 61 or over

Please answer the savings question

PR #677, opened 2016-06-21
Online application saving update

colinbruce:
**This should be merged into the savings-update branch before it's merged into master**

Allows the public app to submit updated inputs and process them.

### DIFF
--- a/app/models/online_application.rb
+++ b/app/models/online_application.rb
@@ -3,7 +3,7 @@ class OnlineApplication < ActiveRecord::Base
 
   validates :ni_number, :date_of_birth, :first_name, :last_name, :address,
     :postcode, presence: true
-  validates :married, :threshold_exceeded, :benefits, :refund, :probate, :email_contact,
+  validates :married, :min_threshold_exceeded, :benefits, :refund, :probate, :email_contact,
     :phone_contact, :post_contact, :feedback_opt_in, inclusion: [true, false]
   validates :reference, uniqueness: true
 
--- a/db/migrate/20160620085231_add_new_saving_fields_to_online_application.rb
+++ b/db/migrate/20160620085231_add_new_saving_fields_to_online_application.rb
@@ -1,0 +1,18 @@
+class AddNewSavingFieldsToOnlineApplication < ActiveRecord::Migration
+  def change
+    rename_column :online_applications, :threshold_exceeded, :min_threshold_exceeded
+    add_column :online_applications, :max_threshold_exceeded, :boolean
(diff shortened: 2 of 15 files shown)


app/models/online_application.rb
```diff
@@ -3,7 +3,7 @@ class OnlineApplication < ActiveRecord::Base
 
   validates :ni_number, :date_of_birth, :first_name, :last_name, :address,
     :postcode, presence: true
-  validates :married, :threshold_exceeded, :benefits, :refund, :probate, :email_contact,
+  validates :married, :min_threshold_exceeded, :benefits, :refund, :probate, :email_contact,
```
novotnyjakub: Should we add here the max threshold as well, to make sure it's `true`/`false`?

db/migrate/20160620085231_add_new_saving_fields_to_online_application.rb
```diff
@@ -0,0 +1,18 @@
...
+    rename_column :online_applications, :threshold_exceeded, :min_threshold_exceeded
+    add_column :online_applications, :max_threshold_exceeded, :boolean
+    add_column :online_applications, :over_61, :boolean
+    add_column :online_applications, :amount, :integer
+
+    reversible do |dir|
+      dir.up do
+        migrate_sql = <<~SQL
+          UPDATE online_applications
+          SET max_threshold_exceeded = min_threshold_exceeded
```
novotnyjakub: I'm not sure about this, because if the `min_threshold_exceeded == false`, then this sets `max_threshold_exceeded = false` whereas I assumed it would be `nil` in that case. Maybe not a big deal.

spec/controllers/online_applications_controller_spec.rb
```diff
@@ -157,7 +156,7 @@
       allow(application).to receive(:save)
       allow(ApplicationCalculation).to receive(:new).with(application).and_return(application_calculation)
       allow(ResolverService).to receive(:new).with(application, user).and_return(resolver_service)
-
+      allow(application.saving).to receive(:save)
```
novotnyjakub: I think here it would be better only to test that the service is being called correctly, instead of actually calling it. Like the other services, you can do:

```ruby
  # define a dummy service
  let(:savings_service) { double(calculate!: nil) }

  # setup the real call to return the dummy service
  before do
    allow(SavingsPassFailService).to receive(:new).with(application.saving).and_return(savings_service)
  end

  # make sure the service has been called
  it 'runs the savings calculation' do
    expect(savings_service).to have_received(:calculate!)
  end 
```

PR #678, opened 2016-06-22
Fixes and extension of Update Savings

colinbruce:
This PR is designed to allow us to add to the `update-savings` branch while allowing a quicker review process...

Fixes to be implemented
===
- [x] On the application summary view - ensure the max_threshold_exceeded label is only shown when explicitly set
- [x] Ensure the Savings pass/fail is shown on the confirmation page (add .to_s to convert_to_boolean)

Features to be implemented
===
- [x] Remove the savings_investment_valid? from application model and make good (2 view models and duplicate for the migration
- [x] The online application builder is not setting the min_ and max_threshold

Nice to have
===
- [ ] ~~Add the Settings.min/max threshold to the labels on savings view and summary~~

### DIFF
--- a/lib/savings_transformation.rb
+++ b/lib/savings_transformation.rb
@@ -41,14 +41,10 @@ class SavingsTransformation
   end
 
   def savings_investment_valid?(application)
-    result = false
-    if application.threshold_exceeded == false ||
-       (
-         application.threshold_exceeded &&
-         (application.partner_over_61 && application.high_threshold_exceeded == false)
-       )
-      result = true
-    end
-    result
+    application.threshold_exceeded == false ||
+      (
+        application.threshold_exceeded &&
+        (application.partner_over_61? && application.high_threshold_exceeded == false)
+      )
   end
 end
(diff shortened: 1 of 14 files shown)


lib/savings_transformation.rb
```diff
@@ -39,4 +39,16 @@ def someone_over_61?(application)
...
+  def savings_investment_valid?(application)
+    result = false
+    if application.threshold_exceeded == false ||
+       (
+         application.threshold_exceeded &&
+         (application.partner_over_61 && application.high_threshold_exceeded == false)
+       )
+      result = true
+    end
+    result
```
novotnyjakub: This whole method could turn out just to the 'if statement condition' with the same result:

```ruby
application.threshold_exceeded == false ||
  (
    application.threshold_exceeded &&
    (application.partner_over_61 && application.high_threshold_exceeded == false)
  )
```

spec/lib/savings_transformation_spec.rb
```diff
@@ -51,7 +63,8 @@
 
     it 'sets passed to equal savings_and_investments_valid? on the application' do
       Saving.all.each do |saving|
```
novotnyjakub: I wouldn't do the loop here, which uses the `savings_investment_valid?` method, which is 100% the same as the implementation, because that doesn't really test the result. I think it would be better to remove the method from the spec and here to test explicitly each case:

```ruby
  expect(application1.savings.passed).to be true
  expect(application2.savings.passed).to be false
  expect(application3.savings.passed).to be false
  expect(application4.savings.passed).to be true
  expect(application5.savings.passed).to be true
```

Or something like this. It's only 5 applications, so it's fine.

novotnyjakub:
👍  looks good now!

PR #679, opened 2016-06-22
Refactor saving validation

colinbruce:
The original validation left something to be desired.

This tightens it up and makes it, hopefully, more refined!


novotnyjakub:
👍  beautiful.. The form object is now much cleaner!

PR #681, opened 2016-06-23
Update offices for MI report

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/120915287)

The digital and MCTS HQ team needed to be excluded from the Financial MI report

This adds a scope and calls it in the report builder service.

### DIFF
--- a/spec/models/business_entity_spec.rb
+++ b/spec/models/business_entity_spec.rb
@@ -29,6 +29,20 @@ RSpec.describe BusinessEntity, type: :model do
     end
   end
 
+  context 'scopes' do
+    let!(:hmcts)   { create(:office, name: 'HMCTS HQ Team ') }
+    let!(:digital) { create(:office, name: 'Digital') }
+    let!(:bristol) { create(:office, name: 'Bristol') }
+
+    describe 'non_digital' do
+      it 'excludes HQ business entities' do
+        # each office gets 2 business_entities by default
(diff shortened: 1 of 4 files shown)


spec/models/business_entity_spec.rb
```diff
@@ -29,6 +29,20 @@
...
+  context 'scopes' do
+    let!(:hmcts)   { create(:office, name: 'HMCTS HQ Team ') }
+    let!(:digital) { create(:office, name: 'Digital') }
+    let!(:bristol) { create(:office, name: 'Bristol') }
+
+    describe 'non_digital' do
+      it 'excludes HQ business entities' do
+        # each office gets 2 business_entities by default
+        expect(described_class.count).to eql(6)
+        expect(described_class.exclude_hq_teams.count).to eql(2)
```
novotnyjakub: In here I would test that those 2 BE are the one from the Bristol office. To make the test more robust. Because the underlying scope could change to include some other 2 and the test wouldn't break.

colinbruce: See https://github.com/ministryofjustice/fr-staffapp/pull/681/commits/d9e5dbac246cdaac0d65d3581b2b66d65a5b7601#diff-0374931bfbf01014af94179de497b302R44

novotnyjakub:
👍  looks good now!

PR #682, opened 2016-06-23
Remove the applikation namespace

### LINKED ISSUE #530: Remove applikation namespace
There is still a model/views/applikation/result model

Investigate and remove if possible.

colinbruce:
[closes #530]

This was the last hold-out of the Applikation namespace, it was
originally used to prevent clashes with the Application namespace
and has gradually been replaced.

This renames the view model to better match the ApplicationResult
model that it inherits and provides a more descriptive name of
it's use.


novotnyjakub:
👍  long needed :)

PR #684, opened 2016-06-24
Changed amount datatype

colinbruce:
By switching to integer from decimal it doesn't add .0 on re-display
[Pivotal bug ticket](https://www.pivotaltracker.com/story/show/122132311)


novotnyjakub:
👍

PR #687, opened 2016-06-27
don't show children input unless it's needed

axemonkey:
[Pivotal](https://www.pivotaltracker.com/story/show/122129503)

Lets the radio/checkbox module handle show/hide for the children input box, and fixes the case where it would not be hidden on page load.


colinbruce:
👍

PR #688, opened 2016-06-29
Manager override on confirmation page

colinbruce:
[Pivotal ticket](#120851949)

Allows users to override an application that has failed via the benefit route.

Still to do 
===
- [x] Refactor the logic used on the display, possibly into the decision_override view model
- [x] Update the processed applications page to display the new`✓ Passed (by manager's decision)` message
- [x] Increase the number of reason radio buttons (see content task)
- [x] Review the content

### DIFF
--- a/spec/features/applications/confirmation_is_rendered_spec.rb
+++ b/spec/features/applications/confirmation_is_rendered_spec.rb
@@ -29,6 +29,10 @@ RSpec.feature 'Confirmation page', type: :feature do
       scenario 'the remission register right hand guidance is no longer shown' do
         expect(page).to have_no_content 'remission register'
       end
+
+      scenario 'the grant help with fees form is rendered' do
+        expect(page).to have_content 'Grant help with fees'
+      end
     end
 
     context 'after user continues from summary' do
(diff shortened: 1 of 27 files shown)


spec/features/applications/confirmation_is_rendered_spec.rb
```diff
@@ -64,6 +68,50 @@
       scenario 'the income label displays correctly' do
         expect(page).to have_xpath('//div[contains(@class,"summary-result") and contains(@class,"part")]', text: 'Waiting for evidence')
       end
+
+      scenario 'the grant help with fees form is not rendered' do
+        expect(page).not_to have_content 'Grant help with fees'
```
novotnyjakub: Only the usual `.to have_no_content` instead of `not_to have_content` - small thing, which occurs here a few times.

spec/models/decision_override_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it { is_expected.to belong_to(:user) }
+
+  describe 'validations' do
+    it { is_expected.to validate_presence_of(:application) }
+    it { is_expected.to validate_presence_of(:user) }
+
+    describe 'reason' do
+      subject(:override) { build_stubbed :decision_override, reason: reason }
+
+      before { override.valid? }
```
novotnyjakub: The `before` block is not necessary as the `be_valid` matcher will run it.

novotnyjakub:
I'm missing new tests in the `process_controller_spec.rb`, or am I blind?

novotnyjakub:
Generally speaking, the code looks good! I'm not sure if the `ProcessController` is the right place, etc. But we've talked about it quite a lot and as it's not bad, let's do it this way..

novotnyjakub:
👍  looks good now!

PR #690, opened 2016-07-05
exclude branch runner builds from active_job

pwyborn:
Hi @novotnyjakub Can you please check the code to exclude "branch runner" jobs from "active_job".
I was not able to set ACTIVE_JOB_ENABLED to "false" in the branch-runner container


pwyborn:
As discussed - looks good!

PR #697, opened 2016-07-20
Accept income range

novotnyjakub:
This PR makes an addition to the income calculation, which supports min/max income thresholds as well. There is no change to the behaviour of the staff app itself, it's a preparation for the integration with `hwf-publicapp`.


colinbruce:
Looks fine 👍

PR #699, opened 2016-07-25
Variable pagination

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/126793691)

On first load the page defaults to 50 records per-page for processed/deleted application indexes.

This adds a selector for 50/100/500/All to allow users to select their preference, for a single use of the page


novotnyjakub:
👍  looks good!

PR #701, opened 2016-07-26
Refactor the W bug

colinbruce:
Replace the code that highlights outcome statements with code that doesn't highlight all text that starts with W!

### DIFF
--- a/app/helpers/summary_helper.rb
+++ b/app/helpers/summary_helper.rb
@@ -44,15 +44,15 @@ module SummaryHelper
   end
 
   def value_style(value)
-    styles = ['column-two-thirds']
-    case value
-    when /^✓/
-      styles << 'summary-result passed'
-    when /^x/
-      styles << 'summary-result failed'
-    when /^Waiting for/
-      styles << 'summary-result part'
-    end
-    styles
+    ['column-two-thirds'].tap do |styles|
+      case value
+      when /^✓/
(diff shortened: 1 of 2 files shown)


app/helpers/summary_helper.rb
```diff
@@ -44,13 +44,15 @@ def build_data_row(object, field)
...
+    styles = ['column-two-thirds']
+    case value
+    when /^✓/
+      styles << 'summary-result passed'
+    when /^x/
+      styles << 'summary-result failed'
+    when /^Waiting for/
+      styles << 'summary-result part'
+    end
+    styles
```
novotnyjakub: This is perfectly fine, I prefer a `tap` solution though. When the return of the method is only the variable, which is constructed in the method, I think it's cleaner, but I'll leave it completely up to you.

```ruby
['column-two-thirds'].tap do |styles|
  case value
    when /^✓/
      styles << 'summary-result passed'
    when /^x/
      styles << 'summary-result failed'
    when /^Waiting for/
      styles << 'summary-result part'
  end
end
```

If you prefer the current solution, merge in! 👍

novotnyjakub: And actually thinking about it, maybe you need a `.join(' ')` in the end, before you return, to make it a string?

novotnyjakub:
👍

PR #702, opened 2016-07-26
Unicorn killer

colinbruce:
image

## Issue - Memory usage has been increasing over time
Memory usage on the production instances has been increasing over time, it was suggested to try refreshing the unicorn workers to free up memory.

## Solution - Implement a solution to reduce memory usage on prod
Kill unicorn workers when memory and/or processes exceed the thresholds

I found two examples for setting the thresholds [here](http://www.rubydoc.info/gems/unicorn-worker-killer/0.4.4) and [here](https://www.digitalocean.com/community/tutorials/how-to-optimize-unicorn-workers-in-a-ruby-on-rails-app) but either seemed very high or low.  I've split the difference and we can monitor the results and adjust as required.

## Possible followup
Another potential culprit is beaver.  If this does not demonstrably reduce the memory usage over time, we should look at something similar for beaver on the sever.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ gem 'redcarpet'
 
 # Use Unicorn as the app server
 gem 'unicorn'
+gem 'unicorn-worker-killer'
 gem 'logstasher', git: 'https://github.com/shadabahmed/logstasher.git',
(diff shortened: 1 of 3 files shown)


novotnyjakub:
I don't know much about this, but I assume this is a standard practice in using the gem, so let's test it! 👍

PR #707, opened 2016-07-28
Upgrade rails and gems

novotnyjakub:
This is updating to latest possible versions, so when we hand it over to the support team, everything is up to date.


novotnyjakub:
I need to test specifically the upgrade of Devise, because it has been upgraded from **3.x** to **4.x**

colinbruce:
Looking good :+1:

PR #710, opened 2016-08-02
Flag evidence checks

colinbruce:
[Ticket](https://www.pivotaltracker.com/story/show/126320983)

Flag NI numbers that have had a failed or returned evidence check.  
* Automatically re-check on subsequent applications.  
* Count the number of failed applications for MI purposes.
* Clear a flag on completion of a successful evidence check

### DIFF
--- a/app/controllers/evidence_controller.rb
+++ b/app/controllers/evidence_controller.rb
@@ -50,6 +50,7 @@ class EvidenceController < ApplicationController
 
   def summary_save
     ResolverService.new(evidence, current_user).complete
+    process_evidence_check_flag
     redirect_to evidence_confirmation_path
   end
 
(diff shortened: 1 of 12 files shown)


app/controllers/evidence_controller.rb
```diff
@@ -62,7 +63,15 @@ def return_letter
...
 
   def return_application
-    redirect_to root_path if ResolverService.new(evidence, current_user).return
+    if ResolverService.new(evidence, current_user).return
+      process_evidence_check_flag
+      redirect_to root_path
+    end
+  end
+
+  def process_evidence_check_flag
```
novotnyjakub: I would add the fact that the service is called from `summary_save` and `return_application` to the controller spec. It doesn't really change the flow of the controller, but it's important to make sure this line is not deleted.

novotnyjakub:
As discussed, I'd add a `type` to the existing `EvidenceCheck` model, so that we can differentiate why they went created.

novotnyjakub:
Should we maybe have a feature test for this too? To make sure it works end-2-end.

PR #712, opened 2016-08-03
Update public_submission_data

colinbruce:
Ensure data loaded has been completed, not just started.
This was causing issues on the live app when a court had started processing an online_application, but had not completed it.


novotnyjakub:
👍

PR #713, opened 2016-08-03
Change benefit override copy evidence question

novotnyjakub:


joshlow:
@colinbruce - one comment, and it's a little pedantic (don't hate me). Can we add back in the 'the' before 'evidence' so it reads:

Is the evidence for the correct applicant and dated within 3 months of the fee being paid?

colinbruce:
The seems eminently sensible and why I asked for a review... I could sworn I cut and pasted the whole line! 😨

joshlow:
My bad @colinbruce - missed in my original copy - not your ctrl + v skills.

PR #717, opened 2016-08-12
Prevent errors when processing online applications

colinbruce:
## Issue
A situation (now rectified) on the public app meant that users could submit applications with incomplete income data.  When a user entered the reference number, edit the application to add fee, date received, etc; but then, when they pushed next, they would get a blank white error screen.

## Fix
This fix means that now, when a user searches for a reference of an affected application, a message is displayed explaining why they cannot proceed.
image

## Incomplete tasks
- [x] Content review of the error message
- [ ] Technical review

### DIFF
--- a/app/services/application_search.rb
+++ b/app/services/application_search.rb
@@ -9,7 +9,7 @@ class ApplicationSearch
 
   def online
     return if !prepare_reference! || application_exists_and_user_can_access ||
-              application_exists_and_user_cannot_access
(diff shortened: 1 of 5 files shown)


joshlow:
Hey @colinbruce, just a small tweak to say why the application can't be processed.

How about the following:

"This application is incomplete and can’t be processed. Please ask the applicant to apply again."

Let me know if okay or your keen eye spots anything?

jsugarman:
👍 looks good

PR #719, opened 2016-08-19
Resolver cost update

colinbruce:
## Problem
When an application is finalised the resolver service is supposed to calculate a final cost to the department.  In some cases this was not completing properly.  In reports, we had applications with a decision of None (no remission, applicant pays full amount) and a decision_cost.

## Reason
When a feature was added to return part-payments and evidence checks, a new outcome type of `return` was created, this was not checked for in the resolver cost calculator

## Fix 
Check for `none` and `return` in the calculator and set the return value to 0 for both.  Update the test scenarios and add a migration script to change the, currently wrong, entries in the DB.

### DIFF
--- a/db/migrate/20160818114238_update_decision_cost.rb
+++ b/db/migrate/20160818114238_update_decision_cost.rb
@@ -1,0 +1,5 @@
+class UpdateDecisionCost < ActiveRecord::Migration
+  def up
+    CorrectReturnedCosts.up!
+  end
+end
--- a/lib/correct_returned_costs.rb
+++ b/lib/correct_returned_costs.rb
@@ -1,0 +1,11 @@
+class CorrectReturnedCosts
+  def self.affected_records
+    Application.where(decision: 'none').where('decision_cost > 0')
+  end
+
+  def self.up!
+    affected_records.each do |application|
+      application.update(decision_cost: 0)
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


lib/correct_returned_costs.rb
```diff
@@ -0,0 +1,11 @@
+class CorrectReturnedCosts
+  def affected_records
+    Application.where(decision: 'none').where('decision_cost > 0')
+  end
+
+  def up!
+    affected_records.each do |application|
+      application.update(decision_cost: 0)
+    end
+  end
```
doctorpod: These could be class methods

db/migrate/20160818114238_update_decision_cost.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateDecisionCost < ActiveRecord::Migration
+  def up
+    CorrectReturnedCosts.new.up!
+  end
```
doctorpod: Could up! be a class method?

spec/lib/correct_returned_costs_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let!(:application1) { create(:application_full_remission, :processed_state, decision_type: 'part', decision_cost: 30, outcome: 'none') }
+  let!(:application2) { create(:application_full_remission, :processed_state, decision_type: 'evidence', decision_cost: 30, outcome: 'none') }
+  let!(:application3) { create(:application_full_remission, :processed_state) }
+
+  describe '#up!' do
+    it 'works in sequence' do
+      expect(affected?).to eql(2)
+      transform.up!
+      expect(affected?).to eql(0)
+    end
```
doctorpod: affected? could be CorrectReturnedCosts.affected_records.size

PR #720, opened 2016-08-19
Online business entities fix

colinbruce:
## Problem
Business entities are not being created for online_applications when they are processed

## Reason
Originally, when adding a paper form, the system created a business entity and reference at the same time and, so, were encapsulated into a single service class.

Since then, we've added a different type of form (online) and, because they already have a reference generated, this meant that business entities were not being created when processing them.

## Fix
Split the service class into two and call the business entity one always and the reference one as needed.  Update the test scenarios and add a migration script to change the, currently wrong, entries in the DB.

### DIFF
--- a/lib/update_missing_business_entities.rb
+++ b/lib/update_missing_business_entities.rb
@@ -1,0 +1,19 @@
+class UpdateMissingBusinessEntities
+
+  def self.affected_records
+    Application.
+      joins(:detail).
+      where(business_entity_id: nil).
+      where.not(
+        reference: nil,
+        office_id: nil,
+        'details.jurisdiction_id': nil
+      )
+  end
+
+  def self.up!
+    affected_records.each do |application|
+      application.update(BusinessEntityGenerator.new(application).attributes)
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


lib/update_missing_business_entities.rb
```diff
@@ -0,0 +1,19 @@
...
+        office_id: nil,
+        'details.jurisdiction_id': nil
+      )
+  end
+
+  def up!
+    affected_records.each do |application|
+      application.update(BusinessEntityGenerator.new(application).attributes)
+    end
+  end
```
doctorpod: Could be class methods

spec/lib/update_missing_business_entities_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:business_entity) { create :business_entity }
+  let(:shared_params) { { office: business_entity.office, jurisdiction: business_entity.jurisdiction } }
+  let(:shared_params_with_business_entity) { { business_entity: business_entity, office: business_entity.office, jurisdiction: business_entity.jurisdiction } }
+  let!(:application1) { create(:application_full_remission, :processed_state, shared_params) }
+  let!(:application2) { create(:application_full_remission, :processed_state, shared_params_with_business_entity) }
+  let!(:application3) { create(:application_full_remission, :processed_state, shared_params) }
+
+  describe '#up!' do
+    it 'works in sequence' do
+      expect(affected?).to eql(2)
```
doctorpod: affected could be UpdateMissingBusinessEntities.affected_records (if you made it a class method)

PR #726, opened 2016-09-16
Part payment should say paid once it has been processed

colinbruce:
[Trello Ticket](https://trello.com/c/1AnbDZWm/12-part-payment-should-say-paid-once-it-is-processed)

When a part payment goes into the list of processed applications it should say "the applicant has paid..."
As a member of staff
I want to see that a part payment has been paid (currently it looks like it is still outstanding)


jsugarman:
👍 looks nice and simple

PR #729, opened 2016-10-03
Added handler for new error type

colinbruce:
A new error is being received from the LAA
This counts and updates the status banner when it occurs


app/services/dwp_monitor.rb
```diff
@@ -25,8 +25,13 @@ def percent
     # When DWP is offline it returns 400 Bad Request
     # maybe extend to search for x00 as first 3 chars to check for 500 errors too
     total = @checks.count.to_f
+    error_total / total * 100.0
```
doctorpod: I always like to include brackets to indicate precedence explicitly - but that's just me

PR #731, opened 2016-10-17
Update npm handling

colinbruce:
* Update the selenium package version
* ensure phantomJS version is always set to 2.1.1
* cache and pre-install in travis

This corrects issued raised by Gemnasium

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
 language: ruby
 sudo: false
-cache: bundler
+cache:
+  bundler: true
+  directories:
+      - travis_phantomjs
+before_install:
+  # Upgrade PhantomJS to v2.1.1.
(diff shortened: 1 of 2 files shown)


jsugarman:
all fine - but might be worth having a chat about this as i had to deal with something similar and ended do this below to rely on travis prebuilt version

```
#Example: prevent PhantomJS auto install, uses PhantomJS already on your path.
# NOTE: travis has phantomjs pre-installed and on path so auto-install
#       not needed (and can intermittently fail with travis success
#       if install site not available)
Jasmine.configure do |config|
  config.prevent_phantom_js_auto_install = true
end
```

colinbruce:
Ticket in backlog to monitor this implementation in line with your comments
https://trello.com/c/U1HyenS5/56-follow-up-jasmine-tests

PR #732, opened 2016-10-17
Remove HTML-safe from I18n

colinbruce:
Replace with _html names that are escaped internally

This removes a brakeman XSS warning


jsugarman:
interesting 👍

PR #733, opened 2016-10-27
Add ET confirmation email

colinbruce:
Add an email for the completion of online ET applications

Duplicated the text from the publicapp confirmation screen, mainly!

### DIFF
--- a/app/views/public_mailer/submission_confirmation_et.html.slim
+++ b/app/views/public_mailer/submission_confirmation_et.html.slim
@@ -1,0 +1,84 @@
+table cellpadding="0" cellspacing="0" border="0" width="600"
+  tr
+    td colspan="2"
+      h2.heading style="margin-bottom: 1.5em; margin-top: 2em;" =t('email.et.greeting')
+
+  tr
+    td colspan="2"
+      p style="margin-bottom: 1.5em; margin-top: 0;" =t('email.et.steps.one_html', reference: @application.reference)
+      p style="margin-bottom: 1.5em; margin-top: 0;" =t('email.et.steps.two')
+
+  tr
+    td colspan="2"
+      table cellpadding="0" cellspacing="0" border="1" bordercolor="#000" width="598"
+        tr
+          td
+            p style="padding: 0 1em 0 1em; margin-bottom: 1em;"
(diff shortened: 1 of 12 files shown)


app/views/public_mailer/submission_confirmation_et.html.slim
```diff
@@ -0,0 +1,84 @@
+table cellpadding="0" cellspacing="0" border="0" width="600"
```
axemonkey: Someone had to do it. HTML email is never pretty, but sometimes it just has to be done.

Good work soldier. Carry on. 👍

spec/mailers/public_mailer_spec.rb
```diff
@@ -34,4 +34,21 @@
       expect(mail.body.encoded).to have_content 'I completed an online application for a court fee refund'
     end
   end
+
+  describe '#et_confirmation' do
+
+    let(:email) { 'foo@bar.com' }
+    let(:mail_data) { create(:online_application, :et, :with_reference, :with_email) }
+    let(:mail) { described_class.submission_confirmation_et(mail_data) }
```
jsugarman: Is deliver_now not needed here? obviously passing tests anyhow

spec/mailers/public_mailer_spec.rb
```diff
@@ -34,4 +34,21 @@
...
+
+    it 'renders the headers' do
+      expect(mail.subject).to eq(I18n.t('email.et.subject'))
+      expect(mail.to).to eq([email])
```
jsugarman: suggestion only: could use include instead of eql to loosen test e.g.
```
include(email)
include('no-reply@helpwithcourtfees.service.gov.uk')
```

jsugarman: couple of comments but all good 👍

spec/mailers/public_mailer_spec.rb
colinbruce: The delivery happens in the MailService class, not in the mailer itself.

spec/mailers/public_mailer_spec.rb
colinbruce: Cool, not seen that pattern before.

PR #734, opened 2016-10-28
Add override to income

colinbruce:
[Trello ticket](https://trello.com/c/L1NWBlRK/10-include-manager-override-question-for-income-applications)

Allow users to override, and grant, applications when they fail income applications too.
Currently override is only allowed on benefit applications.


jsugarman:
👍

PR #735, opened 2016-11-07
Add columns to raw data extract

colinbruce:
[Trello ticket](https://trello.com/c/XWSWvrRm/59-update-raw-data-extract)

Extended the number of columns in the raw data extract

### DIFF
--- a/app/models/views/reports/raw_data_export.rb
+++ b/app/models/views/reports/raw_data_export.rb
@@ -19,7 +19,10 @@ module Views
         married: 'married',
         decision: 'decision',
         amount_to_pay: 'applicant pays',
-        decision_cost: 'departmental cost'
+        decision_cost: 'departmental cost',
+        source: 'source',
+        granted: 'granted?',
+        evidence_checked: 'evidence checked?'
       }.freeze
 
       HEADERS = FIELDS.values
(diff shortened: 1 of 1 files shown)


app/models/views/reports/raw_data_export.rb
```diff
@@ -52,16 +55,34 @@ def data
...
+          business_entities.code AS bec_code,
+          CASE WHEN reference LIKE 'HWF%' THEN 'digital' ELSE 'paper' END AS source,
+          CASE WHEN de.id IS NULL THEN false ELSE true END AS granted,
+          CASE WHEN ec.id IS NULL THEN false ELSE true END AS evidence_checked
+        COLUMNS
+      end
+
+      def joins
+        <<-JOINS
+          LEFT OUTER JOIN offices ON offices.id = applications.office_id
```
jsugarman: I believe there is no difference between left outer join and left join for most, if not all DBs.

jsugarman:
👍 lots of SQL

app/models/views/reports/raw_data_export.rb
colinbruce: Thanks.  Damn it, now I'm showing my age, I'm still writing ANSI92 compliant SQL! :fearful:

PR #736, opened 2016-11-07
General Refactor

colinbruce:
* Update rubocop to 2.3 parser
  * we updated to ruby 2.3 _**5**_ months ago!
  * turned off the Style/FrozenStringLiteralComment as it errored on every single page and is due to be implemented in ruby 3
  * made code-climate happy

* Removed OUTER from the LEFT JOINS
* Updated the recent SQL refactor to use the new, indenting, HEREDOC format

### DIFF
--- a/app/builders/application_builder.rb
+++ b/app/builders/application_builder.rb
@@ -52,7 +52,7 @@ class ApplicationBuilder
   def dependent_attributes(online_application)
     {}.tap do |attributes|
       if online_application.children.present?
-        attributes[:dependents] = online_application.children > 0
+        attributes[:dependents] = online_application.children.positive?
         attributes[:children] = online_application.children
       end
     end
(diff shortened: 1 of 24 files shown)


app/builders/application_builder.rb
```diff
@@ -52,7 +52,7 @@ def online_application_attributes(online_application)
   def dependent_attributes(online_application)
     {}.tap do |attributes|
       if online_application.children.present?
-        attributes[:dependents] = online_application.children > 0
+        attributes[:dependents] = online_application.children.positive?
```
jsugarman: cool

app/models/views/reports/raw_data_export.rb
```diff
@@ -65,7 +65,7 @@ def build_data
       end
 
       def named_columns
-        <<-COLUMNS
+        <<~COLUMNS
```
jsugarman: cool, I is learning

jsugarman:
👍 looking good

PR #740, opened 2016-12-07
Set decision_cost when granting application

colinbruce:
When a user manually overrides a no decision, the value was not being set
This PR ensures that the decision_cost is set on manual override and backdates the change to pre-existing granted applications


lgarvey:
@colinbruce looks good!

PR #744, opened 2017-01-12
Migrate identity codes from BE to SOP

colinbruce:
## Background [Trello Ticket](https://trello.com/c/BC3V2a7x/19-business-entity-codes-changing-in-january)
Traditionally HMCTS used a Business Entity code (BE code) to identity unique parts of the business, e.g. Bristol County Court would be BC100, the Family Division of Bristol would be BC101, Magistrates BC102, etc. Note: codes are fictional!

As of early 2017, they plan to move to a new numbering system, this will be an 8 digit identifier, e.g. 10101010.

The BE codes served two purposes within the HWF system, they allowed financial tracking by division and they formed the basis of paper applications processed in courts.  The first paper form processed by Bristol Family would be BC101-16-00001 for example.

It has been decided that after the switchover date the new SOP fomat is not suitable for use as the basis of reference numbers (e.g. 10101010-17-00001 is not very user friendly!) so, after the switchover date, the current format for online applications will be used for both paper and online forms.  This means that there will be a consistent strategy for referencing HWF applications, HWF-XYX-XYX

## Changes
This PR serves two functions:

* it has to ensure that the new SOP codes are displayed on the financial report to allow tracking against either number
* after the switchover date it uses a dramatically different style of allocating reference numbers.

This involves changing the MVC objects directly concerned as well as a number of service classes and view models that have been used to display/manipulate the core objects


spec/features/admin_can_maintain_business_entities_spec.rb
```diff
@@ -5,108 +5,234 @@
...
+        context 'by adding' do
+          before do
+            click_link 'Add'
+            fill_in 'business_entity_name', with: new_description
+            fill_in 'business_entity_sop_code', with: new_sop_code
+            click_button 'Create business entity'
+          end
+
+          scenario 'the index page reflects the update' do
+            # binding.pry
```
zaparka: we don't need this :)

colinbruce: Darn! :)

GitCop:

 * Commit: 6946eace7f0458496225fe46a142c9ef8d38e67d

---

PR #748, opened 2017-02-13
Feature/rst 33 remove bussines code switch

zaparka:
@colinbruce Could please take a look? I refactored some code and also removed be_code index from db (it was failing when empty).

https://triadmoj.atlassian.net/browse/RST-33


spec/controllers/business_entities_controller_spec.rb
```diff
@@ -1,189 +1,10 @@
...
-        post :confirm_deactivate, office_id: office.id, id: business_entity.id
-      end
-
-      it { is_expected.to have_http_status(:redirect) }
-
-      it { is_expected.to redirect_to(office_business_entities_path) }
-    end
-  end
-
   context 'after the BEC-SOP switchover date' do
```
colinbruce: These context blocks aren't strictly needed any more

zaparka: ok. I will remove them too :)

PR #749, opened 2017-03-10
Bugfix/rst 97 reference duplication

zaparka:
https://triadmoj.atlassian.net/browse/RST-97
The issue was with duplicated reference numbers. SQL query was cached so it returned same results which meant that the reference number didn't change. I was unable to write a test for this that would allow SQL caching in Rspec. I also add ! to update so when there is an error next it will raise an exception. The exception is caught in a controller which is calling this method and a proper error message is displayed.

### DIFF
--- a/app/controllers/applications/process_controller.rb
+++ b/app/controllers/applications/process_controller.rb
@@ -114,6 +114,11 @@ module Applications
     def summary_save
       ResolverService.new(application, current_user).complete
       redirect_to application_confirmation_path(application.id)
+    rescue ActiveRecord::RecordInvalid => ex
+      flash[:alert] = I18n.t('error_messages.summary.validation')
+      Raven.capture_exception(ex, application_id: application.id)
+
+      redirect_to :back
     end
 
     def confirmation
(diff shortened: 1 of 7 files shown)


spec/controllers/applications/process_controller_spec.rb
```diff
@@ -444,25 +444,54 @@
     let(:application) { create :application_full_remission, office: user.office }
     let(:resolver) { instance_double(ResolverService, complete: nil) }
 
-    before do
-      allow(ResolverService).to receive(:new).with(application, user).and_return(resolver)
+    context 'sucess' do
```
leoapost: there's a typo here

app/controllers/applications/process_controller.rb
```diff
@@ -114,6 +114,11 @@ def summary
     def summary_save
       ResolverService.new(application, current_user).complete
       redirect_to application_confirmation_path(application.id)
+    rescue ActiveRecord::RecordInvalid => ex
+      flash[:alert] = I18n.t('error_messages.summary.validation')
+      Raven.capture_exception(ex, application_id: application.id)
+
+      redirect_to :back
```
leoapost: I see in the specs we [expect](https://github.com/ministryofjustice/fr-staffapp/blob/a9f016e091a9864b8651ad1b2701a95476d7d9ad/spec/controllers/applications/process_controller_spec.rb#L488) the user to get redirected to the root url. Maybe it would be better to do the same here as well because `:back` could be anything.

zaparka:
@leoapost Rubocop and your comments done.

PR #750, opened 2017-03-15
update ET confirmation email body texts

leoapost:


leoapost:
@zaparka yes, here: https://github.com/ministryofjustice/fr-staffapp/blob/master/spec/mailers/public_mailer_spec.rb#L38-L52

zaparka:
@leoapost ok. can you write an email test then ?

leoapost:
I don't think we need tests for the email body texts. All we need is to make sure that the correct emails are getting triggered.

PR #755, opened 2017-04-05
Adding sort to processed applications page

zaparka:
Adding simple sort options to processed page.
https://triadmoj.atlassian.net/browse/RST-175


leoapost:
these commits are already included on #756

PR #757, opened 2017-04-06
Feature/rst 170 search

zaparka:
https://triadmoj.atlassian.net/browse/RST-170
Originally I separated the search into another controller, but the shared view is building URLs based on active controller. Which didn't work for search controller without creating duplicated views. So I decide to do a simple version.

### DIFF
--- a/app/views/shared/processed_and_deleted/_per_page_display.html.slim
+++ b/app/views/shared/processed_and_deleted/_per_page_display.html.slim
@@ -7,24 +7,25 @@ ul.pagination.inline-list.util_mt-medium.util_mb-small
       -else
         =link_to(new_per_page, "/#{controller.controller_name}?per_page=#{new_per_page}")
 
-  -if footer
-    li.anchor
-      = link_to('Top of page', "#head")
+-if footer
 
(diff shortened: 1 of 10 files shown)


app/views/shared/processed_and_deleted/_per_page_display.html.slim
```diff
@@ -7,24 +7,25 @@ ul.pagination.inline-list.util_mt-medium.util_mb-small
...
-  li
-    =link_to('DESC', "/#{controller.controller_name}?sort=processed_desc")
-  li
-    span Fee &nbsp;
-    =link_to('ASC', "/#{controller.controller_name}?sort=fee_asc")
-  li
-    =link_to('DESC', "/#{controller.controller_name}?sort=fee_desc")
+    = link_to('Top of page', "#head", class:'anchor')
+
+- unless footer
```
leoapost: how about to change this unless to an `else`?

e.g.
```
if footer
  = link_to('Top of page', "#head", class:'anchor')
else
  ...
end
```

zaparka: Well spotted.

zaparka:
@leoapost fixed.

PR #763, opened 2017-05-03
Fixing IE8 issue of not accepting post method from link_to.

zaparka:
Fix for https://triadmoj.atlassian.net/browse/RST-209


AlexaVB:
All good from my end @zaparka :)

PR #765, opened 2017-05-10
feature/rst-228 inactive staff filters

leoapost:


app/services/user_filters.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  private
+
+  def office(value)
+    @users = @users.where('office_id = ?', value)
+  end
+
+  def activity(value)
+    @users =
```
zaparka: Can you refactor this into something like @users.send(value) ?

leoapost: I'm on it

PR #766, opened 2017-05-11
add link in guides to complete staff guidance

leoapost:


zaparka:
@leoapost Any tests for this?

leoapost:
@zaparka there will be cucumber tests at some point and I don't think this commit is that important to add capybara tests until then.

/cc @AlexaVB

zaparka:
ok

PR #771, opened 2017-05-23
feature/rst 298 staff notifications

leoapost:


db/schema.rb
```diff
@@ -11,11 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170213153644) do
+ActiveRecord::Schema.define(version: 20170518145329) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "tablefunc"
```
zaparka: @leoapost I assume that you check that this is not needed anymore.

leoapost: actually that was auto-removed when I ran the migrations. Let me double check it.

PR #786, opened 2017-08-21
Upgrade rubocop and fix offenses

leoapost:


app/builders/application_builder.rb
```diff
@@ -59,18 +61,21 @@ def dependent_attributes(online_application)
   end
 
   def online_applicant_attributes(online_application)
-    fields = %i[title first_name last_name date_of_birth ni_number married]
+    fields = [:title, :first_name, :last_name, :date_of_birth, :ni_number, :married]
```
zaparka: @leoapost did the new specification told you to do this or did you change the config for this? I thought that we agreed on the style before?

leoapost: I know we discussed this before but I would like to put it on the table for one more time and get your opinions again.

Some of my arguments about the % Notation are:
1. They are not intuitive to every Ruby dev
2. They are hard to understand from the first sight

> When I see `[:foo, :bar]` or `['foo', 'bar']`, I understand exactly what they are.  
> 
> But when I see the following, I spend a sec or two to think.
> `%i[foo bar]` or `%I[foo bar]`
> `%w[foo bar]` or `%W[foo bar]`
> `%q[foo bar]` or `%Q[foo bar]`
> `%s[foo bar]`
> 
3. Rubocop is promoting [different type of braces](https://github.com/bbatsov/ruby-style-guide#percent-literal-braces) for each one of them. I understand the reasons but it looks inconsistent and make the code even more difficult for new Ruby devs.
4. They are error prone 

> e.g. changing `%I` to `%i` will produce different results
> ```
> foo = "hello there"
> %I[foo bar #{foo}]
> => [:"1,1", :b, :"hello there"]
> %i[foo bar #{foo}]
> => [:"1,1", :b, :"\#{foo}"]
> ```

Also a typo can produce unexpected results without being easy to spot 
> Desired results: [:foo, :bar] 
> ```
> %i[foo,bar]       #the comma is the typo
> => [:"foo,bar"]   #actual results. The linter and interpreter won't complain, the app will blow up.
> ```
---
I spent sometime to see what others prefer but I couldn't find any strong arguments. But, here is a list with famous projects and their preference:

% Notation
* Rails
* Discourse
* Spree 

Non % Notation
* Jekyll
* Gitlab
* Capybara
* fastlane
* paperclip

zaparka: Ok. I personally prefer the old way you have some good points there. The reason why we decided to do it new way was Rubocop so not really convincing argument :) So if @garytaylor thinks that the old way ([:abc]) is better and we should keep it I happy to change Rubocop config.

garytaylor: I much prefer an array of symbols rather than %i etc.. - I hate them - they remind me of writing C !

zaparka: Ok cool, lets use symbols :)

leoapost: @garytaylor I don't mind C but I think the % notation in Ruby was inspired from Perl which makes it even worse to me :)

So, if everyone is happy I will update the rubocop config to disable all the % notations and run the `rubocop auto fix` to update the rest of the code.

PR #787, opened 2017-09-11
Adding confirmation when user is deleted.

zaparka:


spec/rails_helper.rb
```diff
@@ -102,6 +102,8 @@
     config.block_unknown_urls
   end
 
+  Capybara.javascript_driver = :webkit
```
garytaylor: Is this going to be suitable for ALL js tests ?  Be careful you don't go writing tests that will only work for this and then you need to change it because it isn't good enough (its not a true webkit browser - more of a port of one in QT so it as some quirks - I never use it personally - I always use real browsers or if I am desperate for speed then use phantomjs)

zaparka: @garytaylor all tests are passing and there is barely any JS in admin. So it should be fine.

PR #788, opened 2017-09-18
Removing codeclimate token from travis.yml

zaparka:


.travis.yml
```diff
@@ -20,8 +20,6 @@ addons:
     - libqt4-dev
     - xvfb
   postgresql: 9.3
-  code_climate:
```
leoapost: are we sure this will not remove the addon?

zaparka: @leoapost nope. as you can see in Travis for this PR the addon is still there and reporting.

PR #790, opened 2017-10-02
Feature/rst 529 list of users appliations

zaparka:
This is solving the issue for users when they forget to write down the reference number of the application that they processed. It also exposes 'open' applications that user 'abandoned'.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -5,6 +5,7 @@ class HomeController < ApplicationController
     manager_setup_progress
     load_graphs_for_admin
     load_waiting_applications
+    load_users_last_applications
     @online_search_form = Forms::Search.new
     @completed_search_form = Forms::Search.new
     @state = DwpMonitor.new.state
@@ -48,6 +49,10 @@ class HomeController < ApplicationController
     end
   end
 
+  def load_users_last_applications
+    @last_updated_applications = last_updated
+  end
+
   def assign_waiting_for_evidence
     @waiting_for_evidence = waiting_for_evidence.map do |application|
       Views::ApplicationList.new(application.evidence_check)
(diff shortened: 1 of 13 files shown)


app/helpers/home_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+  def waiting_for_evidence_path(application)
+    record = Views::ApplicationList.new(application.evidence_check)
+    evidence_path(record.evidence_or_part_payment)
+  end
+
+  def waiting_for_part_payment(application)
```
leoapost: this method has the same code with `waiting_for_evidence_path`. Can we dry it out?

app/models/query/last_updated_applications.rb
```diff
@@ -0,0 +1,11 @@
+module Query
+  class LastUpdatedApplications
+    def initialize(user)
+      @user = user
+    end
+
+    def find
+      @user.applications.order(updated_at: :desc).limit(5)
```
leoapost: This `.limit()` may cause confusion when using this class. 

If it is necessary then how about something like this:
```
def find(limit: nil)
  @user.applications.order(updated_at: :desc).limit(limit)
end
```
and then 
- `Query::LastUpdatedApplications.new.find` will return all the last updated applications
- `Query::LastUpdatedApplications.new.find(limit: 10)` will return the last 10 updated applications

app/helpers/home_helper.rb
zaparka: well spotted.

app/models/query/last_updated_applications.rb
zaparka: yeah make sense

app/controllers/home_controller.rb
```diff
@@ -48,6 +49,10 @@ def load_waiting_applications
     end
   end
 
+  def load_users_last_applications
+    @last_updated_applications = Query::LastUpdatedApplications.new(current_user).find(limit: 5)
```
leoapost: I forgot to mention we may also need to memoize the `@last_updated_applications`.

zaparka: sure

PR #793, opened 2017-10-18
Displaying evidence income on processed application page.

zaparka:


app/models/views/overview/application.rb
```diff
@@ -46,6 +46,11 @@ def total_monthly_income
         @application.income ? format_currency(@application.income.round) : format_threshold_income
       end
 
+      def total_monthly_income_from_evidence
+        return nil if @application.evidence_check.blank?
+        format_currency(@application.evidence_check.income.round)
```
leoapost: I wonder why the income amount needs to be rounded?

zaparka: @leoapost I guess they don't want to deal with nickles.

app/views/shared/processed_and_deleted/_show.html.slim
```diff
@@ -7,7 +7,7 @@ header
 =build_section_with_defaults 'Personal details', @applicant
 =build_section_with_defaults 'Application details', @details
 =build_section 'Benefits', @application_view, %w[benefits]
-=build_section 'Income', @application_view, %w[number_of_children total_monthly_income]
+=build_section 'Income', @application_view, %w[number_of_children total_monthly_income total_monthly_income_from_evidence]
```
garytaylor: I thought we all hated %w's ?

zaparka: @garytaylor I guess this was not refactored. I can create a ticket for refactoring these.

PR #795, opened 2017-10-20
Converting %w to an array definition style

zaparka:


lib/log_stuff.rb
```diff
@@ -65,7 +65,7 @@ def self.compile_fields_and_tags(args)
     return local_fields, local_tags
   end
 
-  %w( fatal error warn info debug ).each do |severity|
+  ["fatal", "error", "warn", "info", "debug"].each do |severity|
```
leoapost: how about to also use single quotes like the other changes?

zaparka: well spotted and done.

PR #799, opened 2017-11-02
Feature/rst 617 healthcheck notification

zaparka:


app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,12 @@
 class ApplicationMailer < ActionMailer::Base
   default from: Settings.mail.from
   layout 'mailer'
+
+  def dwp_is_down_notifier
+    to = Settings.mail.dwp_notification_alert
+    mail(to: to, subject: t('email.dwp_alert_notification..subject')) do |format|
+      format.html { render plain: t('email.dwp_alert_notification.message') }
```
garytaylor: is this ok ? rendering plain text in html email ?

zaparka: @garytaylor I was thinking about it but I couldn't find any reason why this should be an issue.

leoapost: do we need the html format?

leoapost:
Do we know what's the frequency of the notification when DWP is down?

zaparka:
@leoapost we have to set up a cron job for this so it's up to us how often we call this.

app/mailers/application_mailer.rb
zaparka: @leoapost maybe we don't it theory it should be fine to render text only.

PR #800, opened 2017-11-06
Using ENV var for list of DPW emails for notification.

zaparka:


.env.test
```diff
@@ -1,3 +1,4 @@
 REDIRECTION_DOMAINS=trial.domain:production.domain
 ACTIVE_JOB_ENABLED=false
 PROCESSED_DELETED_PER_PAGE=2
+DWP_NOTIFICATION_ALERT_EMAILS=['dan@test.com', 'petr@test.gov.uk']
```
garytaylor: I always think its good practice to freeze array constants - otherwise they are not as constant as you think.

config/settings.yml
```diff
@@ -6,10 +6,7 @@ mail:
   et:
     england_wales: ETHelpwithfees@hmcts.gsi.gov.uk
     scotland: GLASGOWET@hmcts.gsi.gov.uk
-  dwp_notification_alert:
-    - dan.thompson@digital.justice.gov.uk
-    - kathryn.kram@justice.gov.uk
-    - sustainingteamsupport@hmcts.net
+  dwp_notification_alert: <%= ENV['DWP_NOTIFICATION_ALERT_EMAILS'] %>
```
garytaylor: What happens if a developer does not have this set ?
Also,  assuming we are using a .env file, should we not add it to there as well ?

.env.test
zaparka: This is only for a test so I don't really worry about it.

config/settings.yml
zaparka: @garytaylor this is called by rake task so nothing happens. We are setting this value in deploy repo so it's with other ENV values.

PR #802, opened 2017-11-20
Allows flexible configuration of database via ENV vars 

garytaylor:
Allows flexible configuration of database via ENV vars (for development and test environments) for docker configs
If ENV vars are not set, works as before

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -243,7 +243,7 @@ ActiveRecord::Schema.define(version: 20171030112757) do
     t.integer  "income"
     t.boolean  "refund",                        null: false
     t.date     "date_fee_paid"
-    t.boolean  "probate",                       null: false
+    t.boolean  "probate"
     t.string   "deceased_name"
     t.date     "date_of_death"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -243,7 +243,7 @@
     t.integer  "income"
     t.boolean  "refund",                        null: false
     t.date     "date_fee_paid"
-    t.boolean  "probate"
+    t.boolean  "probate",                       null: false
```
leoapost: is this going to cause any issues?

garytaylor: Hmm, I never noticed that - I never changed it - my guess is that the schema.rb in master is wrong as I built this db from scratch.  Ill double check what migrations say

garytaylor: According to the migrations, this is correct - see db/migrate/20160308155249_add_online_application_tables.rb

leoapost: yes, that looks to be the case. Probably worth moving it to a separated commit?

PR #803, opened 2017-11-21
Adding environment value to subject of alert email.

zaparka:
So we know where the alert email is from.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -4,8 +4,9 @@ class ApplicationMailer < ActionMailer::Base
 
   def dwp_is_down_notifier
     to = Settings.mail.dwp_notification_alert
-    mail(to: to, subject: t('email.dwp_alert_notification..subject')) do |format|
-      format.text { render plain: t('email.dwp_alert_notification.message') }
-    end
+    mail(to: to, subject: t('email.dwp_alert_notification.subject',
+      environment: ENV['ENV'])) do |format|
+        format.text { render plain: t('email.dwp_alert_notification.message') }
+      end
   end
 end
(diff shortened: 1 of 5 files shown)


app/mailers/application_mailer.rb
```diff
@@ -4,8 +4,9 @@ class ApplicationMailer < ActionMailer::Base
 
   def dwp_is_down_notifier
     to = Settings.mail.dwp_notification_alert
-    mail(to: to, subject: t('email.dwp_alert_notification..subject')) do |format|
-      format.text { render plain: t('email.dwp_alert_notification.message') }
-    end
+    mail(to: to, subject: t('email.dwp_alert_notification.subject',
+      environment: ENV['ENV'])) do |format|
```
garytaylor: Why have an environment variable called 'ENV' ?  Or am I misreading this ?  Its just not descriptive at all ?

config/locales/emails_en.yml
```diff
@@ -67,5 +67,5 @@ en-GB:
           two: 2. The Central Office will let you know the outcome of your help with fees application, or if they need more information from you. This usually takes no longer than 5 working days.
       warning: 'Your case will not proceed until you send your help with fees reference number and employment tribunal claim number to the tribunal.'
     dwp_alert_notification:
-      subject: Help With Fees - DWP API Checker alert
+      subject: Help With Fees - DWP API Checker alert (%{environment})
```
garytaylor: I assume this environment thing is OK to be nil ?  If someone hasn't set the env var, will things fail ?
If so, we need an example setting in a .env.example or something ?  otherwise where is the definitive list of env vars that we need to spin up the project ?

spec/rails_helper.rb
```diff
@@ -7,6 +7,7 @@
 
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 ENV['RAILS_ENV'] ||= 'test'
+ENV['ENV'] ||= 'test'
```
garytaylor: I don't personally like the idea of modifying ENV - it is supposed to be a representation of the external environment variables.  I know the rspec helper does it - but that is bad - I would not copy that to be honest - we use dotenv for this reason

zaparka:
@garytaylor the reason why I set and use ENV['env'] is to differentiate if the email is sent from the production or staging. we can't use Rails.env because that will be always a production. I can leave it blank for the test but then how do you know that it returns correct value? I can set it just in that test if that would be a better solution?

PR #808, opened 2017-12-06
Feature/rst 594 updating benefit check

zaparka:
We are updating the way how you can tell the app to go through even when a refund is more than 3 months old. We are also simplifying paper evidence check page.

### DIFF
--- a/app/controllers/applications/process_controller.rb
+++ b/app/controllers/applications/process_controller.rb
@@ -34,11 +34,11 @@ module Applications
     end
 
     def application_details_save
-      @form = Forms::Application::Detail.new(application.detail)
-      @form.update_attributes(form_params(:details))
+      app_form_save = ApplicationFormSave.new(application, form_params(:details))
+      @form = app_form_save.details
 
-      if @form.save
-        redirect_to(action: :savings_investments)
+      if app_form_save.success?
+        redirect_to app_form_save.redirect_url
       else
         @jurisdictions = user_jurisdictions
         render :application_details
--- a/app/controllers/applications/process_controller.rb
+++ b/app/controllers/applications/process_controller.rb
@@ -186,7 +186,7 @@ module Applications
       if benefits
         benefit_check_runner.run
         determine_override
-      elsif no_benefits_paper_evidence?
+      elsif benefits && no_benefits_paper_evidence?
         redirect_to application_benefit_override_paper_evidence_path(application)
       else
         redirect_to(action: :income)
--- a/app/controllers/applications/process_controller.rb
+++ b/app/controllers/applications/process_controller.rb
@@ -34,11 +34,11 @@ module Applications
     end
 
     def application_details_save
-      app_form_save = ApplicationFormSave.new(application, form_params(:details))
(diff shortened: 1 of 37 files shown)


app/controllers/applications/process_controller.rb
```diff
@@ -186,6 +186,8 @@ def benefit_check_and_redirect(benefits)
       if benefits
         benefit_check_runner.run
         determine_override
+      elsif no_benefits_paper_evidence?
```
garytaylor: I see you have created an ApplicationFormSave class to tidy up some of the controller (looks more like a repository to me), but then we have this business logic in the controller that has been added to.  Could this not do with a dose of refactoring (not necessarily in this PR) to get the controller being super slim and just providing the http API to the underlying services ?

db/migrate/20171127154854_add_discretion_applied_to_application.rb
```diff
@@ -0,0 +1,9 @@
+class AddDiscretionAppliedToApplication < ActiveRecord::Migration
+  def up
+    add_column :details, :discretion_applied, :boolean
```
leoapost: Should this have default value? Maybe to avoid null values?

zaparka: @leoapost nil means that there was no discretion applied. true means it was granted and false it was denied. so you need to test 3 possibilities.

app/repositories/application_form_repository.rb
```diff
@@ -0,0 +1,52 @@
...
+    "Forms::Application::#{form_name}".constantize
+  end
+
+  def update_form_attributes_and_save
+    @form.update_attributes(form_params)
+    @success = @form.save
+  end
+
+  def continue_with_discretion_applied?
+    @form.discretion_applied != false
```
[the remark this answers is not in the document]
zaparka: @leoapost because nil is an option too and I'm testing if the discretion was not denied aka blank or granted.

app/repositories/application_form_repository.rb
```diff
@@ -0,0 +1,52 @@
...
+  def application_outcome(outcome)
+    application.update(outcome: outcome)
+  end
+
+  def load_template_path
+    if @form.errors.blank? && continue_with_discretion_applied?
+      @redirect_url = application_savings_investments_path(application)
+    elsif @form.errors.blank? && !continue_with_discretion_applied?
+      @redirect_url = application_summary_path(application)
+      application_outcome('none')
```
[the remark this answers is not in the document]
zaparka: sure make sense.

db/migrate/20171127154854_add_discretion_applied_to_application.rb
leoapost: oh, ok. Thanks for the clarification.

app/repositories/application_form_repository.rb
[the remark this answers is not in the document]
zaparka: @leoapost funny enough I had it named set_template_path but the rubocop doesn't like the name.

PR #814, opened 2018-02-23
Displaying admin notification before login too.

zaparka:


app/controllers/notifications_controller.rb
```diff
@@ -19,7 +19,7 @@ def update
   private
 
   def notification
-    @notification = Notification.first
+    @notification = Notification.first || Notification.create
```
leoapost: How about we use the `.first_or_create` method instead?

spec/features/admin_can_manage_the_message_info_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    scenario 'can view the message' do
+      visit '/'
+      expect(page).not_to have_content 'Edit banner'
+      expect(page).to have_content 'This is message from admin, hear, hear.'
+    end
+  end
+
+  context 'befor log in' do
```
leoapost: `befor` is misspelled

zaparka: @leoapost fixed :)

app/controllers/notifications_controller.rb
zaparka: yes, cool. I didn't know about that one. I only knew about find_or_create :)

PR #820, opened 2018-04-12
Feature RST-974 test auto

temi77:
Test automation for feature RST-974. PLEASE DO NOT MREGE


AlexaVB:
Some of the white space looks off but other than that it looks okay, I don't have any context so I suggest @zaparka signs this off

PR #827, opened 2018-04-30
Catching and saving Error response message for invalid request

zaparka:
DWP checker returns response codes. We are now saving them so we know why the DWP check failed.


app/services/benefit_check_service.rb
```diff
@@ -42,6 +42,12 @@ def query_proxy_api
     log_error(e.message, 'Unspecified error')
   end
 
+  def query_proxy_api
+    @response = JSON.parse(RestClient.post("#{ENV['DWP_API_PROXY']}/api/benefit_checks", params))
```
garytaylor: What about if the service has a problem and returns something you are not expecting ?  Is it worth checking the response HTTP status first - if it has returned anything abnormal its pointless trying to get JSON to parse it

spec/services/benefit_check_service_spec.rb
```diff
@@ -57,6 +57,24 @@
...
+      context 'simulating a 400 error' do
+        let(:message) { { 'error': "LSCBC210: Error in request parameter 'Surname'" }.to_json }
+
+        before do
+          stub_request(:post, "#{ENV['DWP_API_PROXY']}/api/benefit_checks").
+            to_return(status: 400, body: message, headers: {})
+          described_class.new(check)
+        end
+
+        it 'returns the error in message' do
```
garytaylor: Maybe even worth simulating a 500 response in here with content you dont care about and probably even in HTML instead of JSON ?  This will prove that a 500 response (applies to others as well) will not break your code and will give you the error you want within your app ?

zaparka: There is a test covering the 500 already :)

app/services/benefit_check_service.rb
zaparka: In the process_proxy_api_call is rescue block that will catch the error thrown by RestClient so it will not get to parsing JSON.

PR #830, opened 2018-05-04
Feature/rst 993 processing query

zaparka:
This is just a Class for exporting some data from HwF admin.


lib/processing_performance_export.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  def check_date_range
+    return if @from.respond_to?(:strftime) &&
+              @to.respond_to?(:strftime)
+    @from = DateTime.parse('May 1 2017 00:00').utc
+    @to = DateTime.parse('April 30 2018 23:59').utc
+  end
+
+  # rubocop:disable all
+  def preformate_data
```
garytaylor: Mis spelt !!

leoapost:
@zaparka I see RST-993 is closed. Do we still need this PR?

zaparka:
@leoapost  I don think we do.

PR #833, opened 2018-05-15
RST-1111 and additional tests for RST-985 and RST-987

zaparka:


zaparka:
This was already approved when merged to EV check branch. so I will merge it

PR #838, opened 2018-07-17
Adding overwrite for DWP message

zaparka:
We can now choose if we want to use default DWP monitor or just say that DWP is down/up.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
 
   resource :notifications, only: [:edit, :update]
+  resource :dwp_warnings
 
   post 'api/submissions' => 'api/submissions#create'
   get 'reports' => 'reports#index'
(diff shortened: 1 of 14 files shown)


config/routes.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
 
   resource :notifications, only: [:edit, :update]
+  resource :dwp_warnings
```
leoapost: Can we limit this to only edit and update?

zaparka: yes, I forget about that one :)

leoapost:
How about the `/healthcheck` endpoint? Don't we want to override that as well?

 ([source code](https://github.com/ministryofjustice/fr-staffapp/blob/master/app/services/health_status.rb#L12))

zaparka:
@leoapost I think that the ```healthcheck``` should reflect the normal check. This is a feature that is aimed at users, not Us.

PR #840, opened 2018-08-13
Prevent managers from changing office details

leoapost:
This PR restricts users with Manager role from editing offices and their jurisdictions and displays a message to contact the Admin if changes needed.


leoapost:
This feature breaks the user setup process for managers. Needs re-thinking.

PR #853, opened 2018-11-21
RST-1421-Updating the dwp_result based on the response code

zaparka:
When the BadRequest is a response from DWP and the response code is LSCBC9xx then the values stored should be "Unspecified error". Otherwise, it's "BadRequest".


leoapost:
@zaparka is this PR still relevant? Can we close it?

zaparka:
@leoapost  I think stakeholders are not interested in this bit anymore. Let's close it.

PR #855, opened 2018-12-04
Moving "Your last applications" at the top of the list

zaparka:
To improve UX I'm moving "Your last applications" to the top of the page so users don't have to scroll all the way down to find it.


AlexaVB:
Approved

PR #856, opened 2018-12-14
rst-1486 cucumber framework

AlexaVB:
Basic Cucumber framework for the HwF's staff app. Aim was to get at least one test passing but I've done a bit more than that. Some feature files are in progress but can be used for manual testing while the steps get defined. See Jira https://tools.hmcts.net/jira/browse/RST-1486


leoapost:
Are the fixtures necessary? The staff app already has [factories](https://github.com/ministryofjustice/fr-staffapp/blob/master/spec/factories).

AlexaVB:
Requires dev time to fix travis error and ignore the codeclimate

zaparka:
This is duplication now the rebased and up to date PR is #860

PR #895, opened 2019-04-15
RST-1772 -Updating copy for evidence checks

zaparka:


AlexaVB:
Looks good but let me try and add a cucumber test or two. Thanks

PR #972, opened 2019-11-14
RST-2361 - Changing dob validation

zaparka:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/RST-2361

### Change description ###
Acceptending under 16 now, but not the ones born in the future.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/models/forms/application/applicant.rb
+++ b/app/models/forms/application/applicant.rb
@@ -92,7 +92,7 @@ module Forms
       end
 
       def too_young?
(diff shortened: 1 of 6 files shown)


AlexaVB:
@zaparka I just approved this but just noticed that we should remove the full stop at the end of the message to keep it consistent with other error messages and to follow GDS

PR #1022, opened 2020-01-17
Updating the frequency query

zaparka:
For CCMCC pilot we counting only against CCMCC applications

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/services/ccmcc_evidence_check_rules.rb
+++ b/app/services/ccmcc_evidence_check_rules.rb
@@ -26,6 +26,10 @@ class CCMCCEvidenceCheckRules
     fee_range_applies?
   end
 
+  def office_id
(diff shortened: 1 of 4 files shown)


app/services/ccmcc_evidence_check_rules.rb
```diff
@@ -26,6 +26,10 @@ def rule_applies?
     fee_range_applies?
   end
 
+  def office_id
+    @office_id ||= Office.find_by(entity_code: OFFICE_CODE).id
```
leoapost: How about we use the `.pluck()` method instead? 

```
Office.where(entity_code: OFFICE_CODE).pluck(:id)
```

I think it will be more efficient.

PR #1363, opened 2022-11-17
Adding cop performance

zaparka:
cop for performance


zaparka:
conflicts with other branch

PR #1864, opened 2024-10-09
RST-6898 Date received limit validation added

eyss1234:


app/models/forms/application/fee_status.rb
```diff
@@ -67,6 +68,19 @@ def min_date
...
 
+      def date_received_within_limit
+        if date_received.present?
+          begin
+            parsed_date = date_received.to_date
+            if parsed_date < 2.years.ago.to_date
+              errors.add(:date_received, "Please enter a valid date in the correct format DD/MM/YYYY")
+            end
+          rescue ArgumentError
+            # If the date is invalid (like day = 32), do nothing; let other validations handle it
```
zaparka, line 79: Could you please add an error message here so it doesn't just catch exception and user wouldn't know what happen. thank you

zaparka (approved):

PR #2541, opened 2026-07-13
RST-8357 - cookie overflow debug

zaparka:
### Jira link
https://tools.hmcts.net/jira/browse/RST-8357


zaparka:
doing different approach

